PR #1, opened 2024-07-22
Fix tests

dkhofer:


src/models.rs
```diff
@@ -313,7 +313,7 @@ impl Path {
             Ok(res) => res,
             Err(rusqlite::Error::QueryReturnedNoRows) => 0,
             Err(_e) => {
-                panic!("something bad happened querying the database")
+                panic!("Error querying the database: {}", _e);
```
Chris7, line 316: can also do `"Error querying the database: {_e}"`

dkhofer: :+1:

PR #3, opened 2024-07-22
Add build/test actions

dkhofer:


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build
```
Chris7, line 19: i think test auto builds

Chris7 (approved):
wondering if we need to explciit build

.github/workflows/test.yml
dkhofer: It does, I autogenerated this file using the github web UI and it included the build part.  Removed

PR #6, opened 2024-07-24
GFA import

dkhofer:


Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 bio = "2.0.0"
 clap = { version = "4.5.8", features = ["derive"] }
+gfa-reader = { git = "https://github.com/MoinSebi/gfa-reader", branch = "master" }
```
Chris7, line 9: can we do a commit hash?

Chris7 (approved):

Cargo.toml
dkhofer: Yep, just added that

src/imports/gfa.rs
```diff
@@ -0,0 +1,197 @@
...
+    let gfa: Gfa<u64, (), ()> = Gfa::parse_gfa_file(gfa_path);
+    let mut sequences_by_segment_id: HashMap<u64, Sequence> = HashMap::new();
+
+    for segment in &gfa.segments {
+        let input_sequence = segment.sequence.get_string(&gfa.sequence);
+        let sequence = Sequence::new()
+            .sequence_type("DNA")
+            .sequence(input_sequence)
+            .save(conn);
+        sequences_by_segment_id.insert(segment.id, sequence);
```
Chris7, line 27: we'll probably need to trim this out at larger scale as we won't be able to store all the sequences in memory

src/imports/gfa.rs
```diff
@@ -0,0 +1,197 @@
...
+
+fn import_gfa(gfa_path: &str, collection_name: &str, conn: &Connection) {
+    models::Collection::create(conn, collection_name);
+    let block_group = BlockGroup::create(conn, collection_name, None, "");
+    let gfa: Gfa<u64, (), ()> = Gfa::parse_gfa_file(gfa_path);
+    let mut sequences_by_segment_id: HashMap<u64, Sequence> = HashMap::new();
+
+    for segment in &gfa.segments {
+        let input_sequence = segment.sequence.get_string(&gfa.sequence);
+        let sequence = Sequence::new()
```
Chris7, line 23: what would a shallow version of this look like?

src/imports/gfa.rs
```diff
@@ -0,0 +1,197 @@
...
+        path_edge_ids.push(edge_id);
+        Path::create(conn, path_name, block_group.id, path_edge_ids);
+    }
+
+    for input_walk in &gfa.walk {
+        let path_name = &input_walk.sample_id;
+        let mut source_hash = Sequence::PATH_START_HASH;
+        let mut source_coordinate = 0;
+        let mut path_edge_ids = vec![];
+        for segment_id in input_walk.walk_id.iter() {
```
Chris7, line 116: does this bring any of the strand info with it? walks do have direction

Chris7 (approved):
just some random comments on things i think we may have to do.

src/imports/gfa.rs
dkhofer: I was being lazy.  I'll check what the parser provides and deal with it in a followup PR

PR #10, opened 2024-08-01
Add code to split blocks and update edges

dkhofer:


migrations/01-initial/up.sql
```diff
@@ -38,7 +38,7 @@ CREATE TABLE block (
   strand TEXT NOT NULL DEFAULT "1",
   FOREIGN KEY(sequence_hash) REFERENCES sequence(hash),
   FOREIGN KEY(block_group_id) REFERENCES block_group(id),
-  constraint chk_strand check (strand in ('-1', '1', '0', '.', '?'))
+  constraint chk_strand check (strand in ('-1', '1', '0', '.', '?', '+', '-'))
```
Chris7, line 41: on the db side, we probably should not use +/- if we have 1/-1

src/models/block.rs
```diff
@@ -58,4 +60,456 @@ impl Block {
...
+    }
+
+    pub fn split(
+        conn: &Connection,
+        block: Block,
+        coordinate: i32,
+        chromosome_index: i32,
+        phased: i32,
+    ) -> Option<(Block, Block)> {
+        if coordinate < block.start || coordinate >= block.end {
```
Chris7, line 125: i think we should panic here instead

migrations/01-initial/up.sql
Chris7: let's actually drop the -1/1/0 for the gff standards

Chris7 (approved):

PR #16, opened 2024-08-19
Alternative to blocks

dkhofer:


src/models/path_edge.rs
```diff
@@ -0,0 +1,243 @@
...
+    #[test]
+    fn test_gets_sequence() {
+        let conn = &mut get_connection();
+        Collection::create(conn, "test collection");
+        let block_group = BlockGroup::create(conn, "test collection", None, "test block group");
+        let sequence1_hash = Sequence::create(conn, "DNA", "ATCGATCG", true);
+        let edge1 = NewEdge::create(
+            conn,
+            NewEdge::PATH_START_HASH.to_string(),
+            -123,
```
Chris7, line 119: why -123 as a coord?

src/models/path.rs
```diff
@@ -124,6 +165,105 @@ impl Path {
...
+
+        let sequence = sequences_by_hash.get(&into.target_hash).unwrap();
+        let start = into.target_coordinate;
+        let end = out_of.source_coordinate;
+
+        let strand;
+        let block_sequence;
+        let block_sequence_length;
+
+        if end >= start {
```
Chris7, line 201: I'm pretty opinionated to not use this convention. I've seen it cause horrors in some dbs that attempted this. We should reserve this for circles where something wraps the origin.

src/models/path_edge.rs
```diff
@@ -0,0 +1,243 @@
+use crate::models::{new_edge::NewEdge, path::Path};
+use rusqlite::types::Value;
+use rusqlite::{params_from_iter, Connection};
+use std::collections::HashMap;
+
+#[derive(Clone, Debug)]
```
Chris7, line 6: I like this, removes the issue of making a graph and finding nodes.

src/models/new_edge.rs
```diff
@@ -0,0 +1,347 @@
+use rusqlite::types::Value;
+use rusqlite::{params_from_iter, Connection};
+use std::collections::{HashMap, HashSet};
+use std::hash::RandomState;
+
+#[derive(Clone, Debug)]
+pub struct NewEdge {
```
Chris7, line 7: I think we should have strand here now, effectively what does the edge point to on the sequence (top side or bottom side)

src/models.rs
```diff
@@ -341,6 +357,200 @@ impl BlockGroup {
...
+            conn,
+            sequence_hashes
+                .into_iter()
+                .map(|hash| format!("\"{hash}\""))
+                .collect(),
+        );
+        let mut blocks = vec![];
+
+        let mut block_index = 0;
+        for (hash, sequence) in sequences_by_hash.into_iter() {
```
Chris7, line 393: does it matter that the blocks aren't going to be in the order of the edges? i don't see anywhere that is keeping order.

src/models.rs
```diff
@@ -341,6 +357,200 @@ impl BlockGroup {
...
+            let sequence_edges = boundary_edges_by_hash.get(&hash);
+            if sequence_edges.is_some() {
+                let sorted_sequence_edges: Vec<NewEdge> = sequence_edges
```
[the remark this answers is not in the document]
Chris7: i think i just need more intuition on boundary edges to comment here. will wait

Chris7 (commented):
Sending back for initial take. Need some clarity on the border stuff.

src/models.rs
```diff
@@ -506,6 +715,112 @@ impl BlockGroup {
...
+                chromosome_index,
+                phased,
+            };
+            new_edges.push(new_start_edge);
+            new_edges.push(new_end_edge);
+        }
+
+        // NOTE: Add edges marking the existing part of the sequence that is being substituted out,
+        // so we can retrieve it as one node of the overall graph
+        if start < start_block.path_end {
```
Chris7, line 794: it seems like start == path_start and end == path_end. should htis code just go in the non-deletion block above?

src/main.rs
```diff
@@ -233,6 +287,162 @@ fn update_with_vcf(
...
+                        let alt_seq = alt_alleles[chromosome_index - 1];
+                        let phased = match gt.phasing {
+                            Phasing::Phased => 1,
+                            Phasing::Unphased => 0,
+                        };
+                        // TODO: new sequence may not be real and be <DEL> or some sort. Handle these.
+                        let new_sequence_hash = Sequence::create(conn, "DNA", alt_seq, true);
+                        let sequences_by_hash = Sequence::sequences_by_hash(
+                            conn,
+                            vec![format!("\"{}\"", new_sequence_hash)],
```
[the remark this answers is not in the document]
dkhofer: 👍 I actually got mad at myself for how I wrote the sequences_by_hash method, I'm going to clean it up in a later PR.  sequences_by_hash should format the strings, not the caller.  I'll use SQLValue::from there

src/models.rs
```diff
@@ -341,6 +357,200 @@ impl BlockGroup {
...
+            if edge.source_hash != NewEdge::PATH_START_HASH {
+                sequence_hashes.insert(edge.source_hash.clone());
+            }
+            if edge.target_hash != NewEdge::PATH_END_HASH {
+                sequence_hashes.insert(edge.target_hash.clone());
+            }
+        }
+
+        let mut boundary_edges_by_hash = HashMap::<String, Vec<NewEdge>>::new();
+        for edge in edges {
```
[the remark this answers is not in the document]
dkhofer: Each represents a split point on an existing sequence, where it's been split into two virtual blocks.

The naive intuition (which I also had) is that if a sequence is split after the bp at index 4, the resulting edge that marks that split should be (4, 5).  But it turns out that if we make it (5, 5), it plays better with the [x, y) notation (0 based end-exclusive) that we're using for blocks, and also the python/rust slice notation.

In fact in general, having the source coordinate of an edge be one greater than the bp index it's referencing works out better, because it fits better with the range indices for blocks.

Let's assume we're using the naive version.  Say we have a pair of edges that define a block:

Edge 1: target sequence S1, target coordinate 5
Edge 2: source sequence S1, source coordinate 10

Then in rust slice terms, the subsequence is S1[5..11].  To get the correct slice (or block) end coordinate, you need to add one to the source coordinate.

Then, in insert_change, we have some blocks and we need to create edges between them.  To create an edge that goes from the end of a range [x, y) to the beginning of a range [z, a), the source coordinate would be y-1.

There are at least 6 lines of code where if we use the naive version, we need to add 1 or subtract 1 to convert from a block/slice coordinate to an edge coordinate or vice versa.  I finally gave up and went with the convention that the edge source coordinate is 1 greater than the bp index it's referencing.

Does that make sense?

src/models.rs
dkhofer: This method is only used for get_all_sequences, so we don't care about ordering the blocks.  We just want to get blocks back that have a way to connect them to the edges, so we can feed them into the graph.  (The way to connect them is, the blocks have a sequence hash and start/end coordinates, and we get the edges that match on target hash/coordinate or source hash/coordinate to know which ones are going into or coming out of a block.)

src/models.rs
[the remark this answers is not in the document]
dkhofer: (It took me a while to get to that concise description.  All I knew when I was trying to get the unit tests passing was that I kept screwing up one coordinate or another, until I realized the edge coordinates were using a different range definition than the block/slice coordinates.)

src/models.rs
dkhofer: There's a related method in path.rs that does do the ordering like you're expecting, because that is needed for a path

src/models.rs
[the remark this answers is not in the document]
dkhofer: See above comments

src/models.rs
dkhofer: Yep, see above.  This is the part of the code that adds the boundary edges.  Intuitively, the conditional is checking if the virtual blocks are already split at this location.  If start == start_block.path_end, that's the case, because the path_end field is the location of the split in the path.

src/models/new_edge.rs
dkhofer: We can get strand for a "virtual block" by looking at the pair of edges that defines the block.  If the first edge's target coordinate is greater than the second edge's source coordinate, the strand is reverse.  See edge_pairs_to_block in path.rs

I know all this is nonintuitive, it took me a while to figure it all out too

src/models/path.rs
dkhofer: Ah, I see.  OK, I can add a strand field to the new edges

src/models/path_edge.rs
dkhofer: Basically anything less than 0 works, this is the input edge from the virtual "start node".  I'm not 100% what we want to do for this, and I don't remember but I think I put -123 because it showed the arbitrariness of how it works.

It turns out that it's useful to have virtual start and end blocks in the interval tree of blocks for a path, so I think we want to keep this around in some form.  Unless this seems seriously bogus to you, I'd like to keep it in this PR and then write up the pros and cons of it and discuss, and if we want to change anything do that next week

migrations/01-initial/up.sql
```diff
@@ -79,4 +79,39 @@ CREATE TABLE change_log (
...
+CREATE TABLE block_group_edges (
+  id INTEGER PRIMARY KEY NOT NULL,
+  block_group_id INTEGER NOT NULL,
+  edge_id INTEGER NOT NULL,
+  FOREIGN KEY(block_group_id) REFERENCES block_group(id),
+  FOREIGN KEY(edge_id) REFERENCES new_edges(id)
+);
+CREATE UNIQUE INDEX block_group_edges_uidx ON block_group_edges(block_group_id, edge_id);
+
+INSERT INTO sequence (hash, sequence_type, sequence, "length") values ("yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "OTHER", "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", 64), ("zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "OTHER", "zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", 64);
```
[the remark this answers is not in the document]
dkhofer: It looked to me like the hashes are hex, so I chose y's and z's because they wouldn't collide.  I might be misunderstanding the suggestion though

src/models.rs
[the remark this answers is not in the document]
dkhofer: The tricky thing to wrap the brain around is that edges are the complement of blocks, so the beginning of an edge is the end of a block, and vice versa, so it's nice to have the coordinate systems agree.

src/models/new_edge.rs
dkhofer: Forget it, saw your later comment, I'll put strand on the edge

migrations/01-initial/up.sql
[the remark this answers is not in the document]
Chris7: You got it right. I didn't realize sha was alphanumerical, but generally we're saying the same thing. I meant start/end just to make it abundantly obvious what this represents when looking at the db. Or edge-start edge-termination or whatever.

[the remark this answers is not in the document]
dkhofer: I renamed them to "start-node-yyyy..." and "end-node-zzzz..." to make it a bit clearer

dkhofer:
Changes today:
* Reworked Sequence::sequences_by_hash() to do the formatting itself.  I couldn't get SQLValue working with a vec of strings and the sql syntax, but the formatting is just in that one method now so callers don't have to worry about it
* Added source_strand and target_strand to new edges, and changed the relevant logic for RC sequence parts

src/models.rs
```diff
@@ -341,6 +357,195 @@ impl BlockGroup {
...
+                let sorted_sequence_edges: Vec<NewEdge> = sequence_edges
+                    .unwrap()
+                    .iter()
+                    .sorted_by(|edge1, edge2| {
+                        Ord::cmp(&edge1.source_coordinate, &edge2.source_coordinate)
+                    })
+                    .cloned()
+                    .collect();
+                let first_edge = sorted_sequence_edges[0].clone();
+                let start = 0;
```
Chris7, line 400: confused why start would always be 0, but i think we're going to be dropping this code anyways to get rid of boundary edges/branch points.

Chris7: like suppose i have a sample where the beginning of chr1 is truncated -- we would still probably be using the same sequence object chr1, but the first start might be like bp 1M.

src/models.rs
```diff
@@ -341,6 +357,195 @@ impl BlockGroup {
...
+                    id: block_index,
+                    sequence_hash: hash.clone(),
+                    sequence: block_sequence,
+                    start,
+                    end,
+                };
+                blocks.push(last_block);
+                block_index += 1;
+            } else {
+                blocks.push(GroupBlock {
```
Chris7, line 441: For future note -- on the builder/default pattern, i think we need to go w/ builder and this makes me realize a great reason why -- shallow sequences will be a real pain to deal with at this level everywhere.

src/models.rs
Chris7: or in other words, start/end always refer to the position in the sequence string, not the coordinates so much. The coordinates are obtained by a collection of blocks w/ lengths.

Chris7 (approved):
i still need to work out some things, but i think this is in a place it should be mergd so we can explore it more.

src/models.rs
dkhofer: The goal with this part of the code is to iterate over all sequences, and for each sequence, to get its blocks, which will then be fed into the graph (in order to get all sequences).  The first block of a sequence will always start with 0, so I don't think this will change.  If it's still not making sense, we can discuss more

src/models.rs
dkhofer: :+1:

PR #17, opened 2024-08-21
Add in simple transactions

Chris7:
Rusqlite's transaction w/ the borrow checker is a real PITA. This is a bandaid (or maybe long term approach) to prevent db corruption for killed processes.


dkhofer (approved):
Sorry for the late review.  LGTM

PR #20, opened 2024-08-26
Switch to noodle fasta parser

Chris7:


dkhofer (approved):
Approving.  Thanks for basing it off my branch!

PR #23, opened 2024-08-27
Cache block group, path, and interval tree for path, and batch inserts

dkhofer:


src/main.rs
```diff
@@ -124,9 +125,87 @@ fn import_fasta(fasta: &String, name: &str, shallow: bool, conn: &mut Connection
...
+    }
+
+    pub fn lookup(
+        block_group_cache: &mut BlockGroupCache,
+        collection_name: String,
+        sample_name: String,
+        name: String,
+    ) -> i32 {
+        let block_group_key = BlockGroupData {
+            collection_name: collection_name.to_string(),
```
Chris7, line 160: this is already a string. I think this block can be made simpler by making the String arguments into &str

src/main.rs
```diff
@@ -224,46 +299,17 @@ fn update_with_vcf(
                                 let allele = allele.unwrap();
                                 if allele != 0 {
                                     let alt_seq = alt_alleles[allele - 1];
-                                    // TODO: new sequence may not be real and be <DEL> or some sort. Handle these.
```
Chris7, line 241: thanks for consolidating this

src/main.rs
```diff
@@ -238,48 +314,17 @@ fn update_with_vcf(
...
-                                        &sample_paths[0],
+                                        &sample_path,
+                                        alt_seq,
                                         ref_start as i32,
                                         ref_end as i32,
-                                        &new_block,
                                         chromosome_index as i32,
                                         phased,
                                     );
+                                    changes.push(change);
```
Chris7, line 327: we should think about checkpointing this to make it inserts of like 10k or 50k to reduce memory pressure. Could get messy with the many gb vcfs as well.

src/main.rs
```diff
@@ -135,9 +136,91 @@ fn import_fasta(fasta: &String, name: &str, shallow: bool, conn: &mut Connection
...
+    ref_end: i32,
+    chromosome_index: i32,
+    phased: i32,
+) -> PathChange {
+    // TODO: new sequence may not be real and be <DEL> or some sort. Handle these.
+    let new_sequence_hash = Sequence::new()
+        .sequence_type("DNA")
+        .sequence(alt_seq)
+        .save(conn);
+    let sequence =
```
Chris7, line 198: This could be a good place to have a small cache. For two cases this has great benefits:
* For many samples, we iterate by samples then by genotype, meaning we can be looking up the sample 3-4 alt alleles many times.
* For trivial changes, every simple SNP is a lookup
 
Maybe a cache size of like 200?

Chris7 (approved):
LGTM. One comment on a place i think we can potentially get a good speedup.

src/main.rs
dkhofer: I converted as many as I could for the block group stuff, but ran in to some reference lifetime issues with one variable, so I'm declaring success

src/main.rs
dkhofer: Definitely, but not going to do it for this PR.  Adding it to the list though

src/main.rs
dkhofer: Agree.  I'm going to do that in my next PR.  I'd like to merge this one

PR #24, opened 2024-08-27
Support sequences stored on disk

Chris7:
Support storing sequences on disk instead of in db.

The builder pattern is used here, as it allows other defaults to be set automatically which was not possible in the default pattern.

Given that the sequence itself needs some inference to know how to fetch it, sequence was made private as well.

### DIFF
--- a/src/models/sequence.rs
+++ b/src/models/sequence.rs
@@ -31,13 +31,6 @@ pub struct NewSequence<'a> {
 }
 
 impl<'a> NewSequence<'a> {
-    pub fn new() -> NewSequence<'static> {
-        NewSequence {
-            shallow: false,
-            ..NewSequence::default()
-        }
-    }
-
     pub fn shallow(mut self, setting: bool) -> Self {
         self.shallow = setting;
         self
@@ -50,6 +43,7 @@ impl<'a> NewSequence<'a> {
 
     pub fn sequence(mut self, sequence: &'a str) -> Self {
         self.sequence = Some(sequence);
+        self.length = Some(sequence.len() as i32);
         self
     }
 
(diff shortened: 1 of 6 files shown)


src/models/sequence.rs
```diff
@@ -1,22 +1,125 @@
...
     pub length: i32,
+    // indicates whether the sequence is stored externally, a quick flag instead of having to
+    // check sequence or file_path and do the logic in function calls.
+    pub external_sequence: bool,
 }
 
-impl Sequence {
-    pub fn create(conn: &Connection, sequence_type: &str, sequence: &str, store: bool) -> String {
+#[derive(Default)]
+pub struct NewSequence<'a> {
```
dkhofer, line 24: What does the `'a` syntax mean?

src/models/sequence.rs
```diff
@@ -87,3 +281,125 @@ impl Sequence {
...
+            .file_path("/foo/bar")
+            .length(50)
+            .build();
+        assert_eq!(sequence.length, 50);
+        assert_eq!(sequence.sequence, "");
+    }
+
+    #[test]
+    fn test_create_sequence_in_db() {
+        let conn = &mut get_connection();
```
dkhofer, line 325: I don't think connection needs to be mutable.  Maybe I'll go through and remove the `mut` from all the instances, so it doesn't get copy/pasted around

dkhofer:
I already was familiar with and liked the builder pattern.  Using it in conjunction with passing in the conn object is great.  LGTM, comments are nonblocking

dkhofer (approved):

src/models/sequence.rs
Chris7: They are for lifetimes. I won't pretend to understand them well atm: https://doc.rust-lang.org/rust-by-example/scope/lifetime.html

PR #26, opened 2024-08-28
Add sequence cache

dkhofer:


src/main.rs
```diff
@@ -179,29 +179,72 @@ impl<'a> BlockGroupCache<'_> {
...
+        SequenceCache {
+            cache: HashMap::<SequenceKey, Sequence>::new(),
+            conn,
+        }
+    }
+
+    pub fn lookup(
+        sequence_cache: &mut SequenceCache<'a>,
+        sequence_type: &'a str,
+        sequence: String,
```
Chris7, line 205: i think this will result in ownership issues where String is moved here but not returned out. &String can resolve that

src/main.rs
```diff
@@ -179,29 +179,72 @@ impl<'a> BlockGroupCache<'_> {
...
+        };
+        let sequence_lookup = sequence_cache.cache.get(&sequence_key);
+        if let Some(found_sequence) = sequence_lookup {
+            found_sequence.clone()
+        } else {
+            let new_sequence_hash = Sequence::new()
+                .sequence_type("DNA")
+                .sequence(&sequence)
+                .save(sequence_cache.conn);
+            let new_sequence =
```
Chris7, line 219: i think you can actually do this w/o the additional lookup by using the `build` part of the builder to get the sequence object. So you would have the builder object as a var, call `save` to get the hash, and `build` to get the sequence object.

Chris7 (approved):
LGTM. Some suggestions on a way to avoid an extra db lookup. I think the lookup use of `String` will lead to a refactor down the line using this method if we're passing a sequence we want ot reuse from the caller.

src/main.rs
dkhofer: I was actually fighting the compiler about this yesterday.  I couldn't get it to work with &str.  I tried with &String just now and got the same problem, which is that the sequence cache exists outside the for loop that's processing the VCF records.  The sequence that's being cached gets created inside that for loop, so any reference to it is transient (could go away from one iteration of the for loop to the next), so the sequence reference can't get added to the cache.

src/main.rs
dkhofer: Good idea, done!

PR #28, opened 2024-08-29
Get rid of boundary edges in the database

dkhofer:


src/main.rs
```diff
@@ -360,6 +360,7 @@ fn main() {
             sample,
         }) => {
             let mut conn = get_connection(db);
+            conn.execute("PRAGMA cache_size=50000;", []).unwrap();
```
dkhofer, line 363: This dropped the benchmark timing from 11s to 8.5-9s

src/models/block_group_edge.rs
```diff
@@ -11,18 +11,21 @@ pub struct BlockGroupEdge {
...
-            formatted_rows_to_insert
-        );
-        let _ = conn.execute(&insert_statement, ());
+            let formatted_rows_to_insert = rows_to_insert.join(", ");
+
+            let insert_statement = format!(
+                "INSERT OR IGNORE INTO block_group_edges (block_group_id, edge_id) VALUES {0};",
+                formatted_rows_to_insert
+            );
+            let _ = conn.execute(&insert_statement, ());
```
Chris7, line 27: do we even need a nassignment?

src/models/block_group.rs
```diff
@@ -197,56 +197,97 @@ impl BlockGroup {
...
+                    target_hash: hash.clone(),
+                    target_coordinate: *block_boundary,
+                    target_strand: String::from(""),
+                    chromosome_index: 0,
+                    phased: 0,
+                });
+            }
+
+            #[allow(clippy::len_zero)]
+            if block_boundaries.len() > 0 {
```
Chris7, line 288: guessing thi swants you to use .empty() instead of len?

src/models/block_group.rs
```diff
@@ -257,9 +298,7 @@ impl BlockGroup {
                 };
                 blocks.push(first_block);
                 block_index += 1;
-                for (into, out_of) in sorted_sequence_edges.clone().into_iter().tuple_windows() {
-                    let start = into.target_coordinate;
-                    let end = out_of.source_coordinate;
+                for (start, end) in block_boundaries.clone().into_iter().tuple_windows() {
```
Chris7, line 301: this is nice to know we have, i was just doing the other way for zipping lists

Chris7 (approved):

src/models/block_group_edge.rs
dkhofer: The rust compiler gives a warning if I don't have it, because `execute` may return an Err value.  It seems like the shortest way of saying "yes, I actually mean to ignore the return value".

src/models/block_group.rs
dkhofer: Yeah, but I just tried again and it worked.  I think I was in a rush and tried `is_empty` instead of `!is_empty`, which caused tests to break.  I think at that point I had been fixing compiler and formatting errors for a while and just got fed up with it.  Thanks for asking about this, it's fixed now

PR #31, opened 2024-09-05
Changelog

Chris7:


src/models/change_log.rs
```diff
@@ -0,0 +1,353 @@
...
+use crate::models::block_group_edge::BlockGroupEdge;
+use crate::models::edge::Edge;
+use crate::models::path::{NewBlock, Path};
+use crate::models::sequence::Sequence;
+use chrono::prelude::*;
+use rusqlite::types::Value;
+use rusqlite::{params_from_iter, Connection};
+
+#[derive(Debug)]
+pub struct ChangeLog {
```
dkhofer, line 11: I just realized this is very similar to the PathChange struct I added to help with bulk creation of edges in insert_changes: https://github.com/ginkgobioworks/gen/blob/main/src/models/block_group.rs#L46

It doesn't need to happen in this PR, but we should probably refactor to consolidate.

Should ChangeLogs have chromosome_index and phased fields?

src/test_helpers.rs
```diff
@@ -0,0 +1,22 @@
+use std::fs;
+
+use rusqlite::Connection;
+
+use crate::migrations::run_migrations;
+
+pub fn get_connection<'a>(db_path: impl Into<Option<&'a str>>) -> Connection {
```
dkhofer, line 7: Sweet, thanks for this!  Would you mind updating the other model files that roll their own `get_connection` to use this as well?

dkhofer:
LGTM, comments are nonblocking

dkhofer (approved):

src/models/change_log.rs
Chris7: So a reason for this was the needs of each struct may evolve independently and i didn't want to encumber pathchange with choices not related to it.

Chris7: i think i likely need those fields though too!

Chris7: hmm, maybe not -- i think the path_id edges should actually give whatever chromosome/phase is relevant

Chris7: i'll think about it and add a note.

PR #32, opened 2024-09-05
Handle edits at beginning or end of path correctly

dkhofer:


src/models/block_group.rs
```diff
@@ -10,7 +10,7 @@ use crate::models::block_group_edge::BlockGroupEdge;
 use crate::models::edge::{Edge, EdgeData};
 use crate::models::path::{NewBlock, Path, PathData};
 use crate::models::path_edge::PathEdge;
-use crate::models::sequence::{NewSequence, Sequence};
```
dkhofer, line 13: Just doing some warning cleanup here and elsewhere

Chris7 (approved):
this is great

PR #35, opened 2024-09-06
Move library functionality out of main

Chris7:
This separates the binary client interface from the library


dkhofer:
Thanks for this, if you hadn't done it, I was going to get it soon.  One suggestion, maybe just have a `src/commands` dir and put both the fasta and vcf stuff under that?

dkhofer (approved):

PR #37, opened 2024-09-11
Handle pooled sequences better

dkhofer:


src/graph.rs
```diff
@@ -40,3 +40,125 @@ where
         None
     })
 }
+
+#[cfg(test)]
```
dkhofer, line 44: My first suspicion was that the graph code wasn't right, so I added some unit tests here.  But they passed, so I realized this code was fine.

src/models/block_group.rs
```diff
@@ -290,31 +293,16 @@ impl BlockGroup {
...
+        let mut sequence_blocks_by_hash = HashMap::new();
         let mut block_index = 0;
-        let mut boundary_edges = vec![];
         for (hash, sequence) in sequences_by_hash.into_iter() {
             let block_boundaries = BlockGroup::get_block_boundaries(
                 edges_by_source_hash.get(hash.as_str()),
                 edges_by_target_hash.get(hash.as_str()),
                 sequence.length,
             );
-            for block_boundary in &block_boundaries {
```
dkhofer, line 302: This method has a lot going on, so I decided to refactor a bit to separate out the boundary edge code

src/models/block_group.rs
```diff
@@ -385,14 +374,124 @@ impl BlockGroup {
...
+                    target_strand: Strand::Unknown,
+                    chromosome_index: 0,
+                    phased: 0,
+                });
+            }
+        }
+        boundary_edges
+    }
+
+    pub fn get_completion_edges(
```
dkhofer, line 405: I'm open to a better name

src/models/block_group.rs
```diff
@@ -385,14 +374,124 @@ impl BlockGroup {
...
     }
 
     pub fn get_all_sequences(conn: &Connection, block_group_id: i32) -> HashSet<String> {
         let mut edges = BlockGroupEdge::edges_for_block_group(conn, block_group_id);
-        let (blocks, boundary_edges) = BlockGroup::blocks_from_edges(conn, &edges);
+        let sequence_blocks_by_hash = BlockGroup::blocks_from_edges(conn, &edges);
+        let boundary_edges =
+            BlockGroup::boundary_edges_from_sequences(sequence_blocks_by_hash.clone());
         edges.extend(boundary_edges.clone());
+        let completion_edges =
```
dkhofer, line 487: So this fixes the bug, but it's only in the context of get_all_sequences.  We will also need to get edges like this for pooled sequence operations, but since it's not clear yet how those operations/commands will work, we will need to wait for their implementation to use this code.

src/models/block_group.rs
Chris7: it seems wrong that we are doing these permutations manually vs. them coming out of the graph naturally. Are we missing a common node between the two regions? Would nodes with sequences of "" solve this w/o special treatment here?

bobvh: One problem with a null-node between each block is that you then force full-combinatorial for everything. Defining allowed sub-paths would help deconvolute, or you could encode associations not in the sequence graph itself but as a special type of edge.

Chris7: I'm not sure why we wouldn't want the full combinatorial. This seems to imply we have some edge group mechanism. Suppose i added sequences x/y/z to the same pool a week later, would i want to now exclude the promoters? Maybe i do, maybe i didn't.

bobvh: How's this for syntax of an update command that handles both: https://github.com/ginkgobioworks/gen/tree/userdocs-outlining?tab=readme-ov-file#recording-sequence-changes ? It covers most of the cases I've seen in practice. With examples it should be more clear in expanded docs.

dkhofer:
Closing this, going to start a new branch with Bob's suggested flags for combinatorial vs non-combinatorial

PR #38, opened 2024-09-11
Add GFA import command, and handle reverse strand segments etc.

dkhofer:


src/main.rs
```diff
@@ -71,13 +75,35 @@ fn main() {
...
             name,
             db,
             shallow,
         }) => {
             let mut conn = get_connection(db);
             conn.execute("BEGIN TRANSACTION", []).unwrap();
-            import_fasta(fasta, name, *shallow, &mut conn);
+            if fasta.is_some() {
+                import_fasta(
+                    &<Option<std::string::String> as Clone>::clone(fasta)
```
Chris7, line 87: curious why we have the other stuff here vs. fasta.unwrap()

Chris7 (approved):
lgtm, just curious on a rust thing.

src/main.rs
dkhofer: I think I was blindly following the advice of the compiler.  It ended up being &fasta.clone().unwrap() because it didn't like me borrowing the value of fasta.unwrap()

PR #40, opened 2024-09-12
Operation tracking

Chris7:
This adds in operations for changes
* All the changelog stuff is gone since it's not being used
* Tracking edges w/ operations is also gone. I have a bad feeling about trying to correlate primary keys that may not align on applications of changes and that it may be a better approach to do something like repeating the actual vcf import/etc. for merging changes across branches.
* How to display changes is still an open question, next PR will be on the command line interface for managing operations.

I don't like how the filesystem management is happening right now, as we have a few scenarios that make me concerned -- having 2 state stores is bad and leads to problems around 2-phase commits and a neverending possibility of disagreeing systems. I'm ok leaving that for future us.


migrations/01-initial/up.sql
```diff
@@ -34,34 +34,27 @@ CREATE TABLE path (
...
 ) STRICT;
 
-CREATE TABLE change_set_changes (
+CREATE TABLE operation_summary (
   id INTEGER PRIMARY KEY NOT NULL,
-  change_set_id INTEGER NOT NULL,
-  change_log_id INTEGER NOT NULL,
-  FOREIGN KEY(change_set_id) REFERENCES change_set(id),
-  FOREIGN KEY(change_log_id) REFERENCES change_log(id)
+  operation_id INTEGER NOT NULL,
```
dkhofer, line 55: Should there be a foreign key constraint for this?

src/operation_management.rs
```diff
@@ -0,0 +1,172 @@
...
+
+pub fn revert_changeset(conn: &Connection, op_id: i32) {
+    let change_path = Path::new(&get_gen_dir()).join(format!("{op_id}.cs"));
+    let mut file = fs::File::open(change_path).unwrap();
+    let mut contents = vec![];
+    file.read_to_end(&mut contents).unwrap();
+    let mut inverted_contents: Vec<u8> = vec![];
+    session::invert_strm(&mut &contents[..], &mut inverted_contents);
+
+    conn.pragma_update(None, "foreign_keys", "0").unwrap();
```
dkhofer, line 79: Might be good to have an error handling block that ensures foreign key checks are put back even if the database statement fails

dkhofer (approved):

migrations/01-initial/up.sql
Chris7: you are right

PR #41, opened 2024-09-13
Add in ui and enable multi-db support of operations

Chris7:
Adds cli interface for working with operations, adds in a unique uuid per database for scoping changes to a given db

### DIFF
--- a/migrations/01-initial/up.sql
+++ b/migrations/01-initial/up.sql
@@ -1,3 +1,7 @@
+CREATE TABLE gen_metadata (
+  db_uuid TEXT PRIMARY KEY NOT NULL
+) STRICT;
+
 CREATE TABLE collection (
   name TEXT PRIMARY KEY NOT NULL
 ) STRICT;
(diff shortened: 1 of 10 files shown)


migrations/01-initial/up.sql
```diff
@@ -93,3 +97,10 @@ CREATE TABLE block_group_edges (
 CREATE UNIQUE INDEX block_group_edges_uidx ON block_group_edges(block_group_id, edge_id);
 
 INSERT INTO sequence (hash, sequence_type, sequence, name, file_path, "length") values ("start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "OTHER", "start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "", "", 64), ("end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "OTHER", "end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "", "", 64);
+INSERT INTO gen_metadata (db_uuid) values (lower(
+    hex(randomblob(4)) || '-' || hex(randomblob(2)) || '-' || '4' ||
```
dkhofer, line 101: Is this like a standard UUID format?

src/operation_management.rs
```diff
@@ -85,6 +89,29 @@ pub fn revert_changeset(conn: &Connection, op_id: i32) {
     conn.pragma_update(None, "foreign_keys", "1").unwrap();
 }
 
+pub fn move_to(conn: &Connection, op_id: i32) {
```
dkhofer, line 92: I don't see any way around this, but I hope we don't end up reinventing a bunch of git internals

dkhofer:
LGTM

dkhofer (approved):

migrations/01-initial/up.sql
Chris7: yeah, we can discuss if we should make this something simpler like the local db path and checksum it or something.

PR #42, opened 2024-09-13
GFA export

dkhofer:


src/models/edge.rs
```diff
@@ -224,6 +241,219 @@ impl Edge {
...
+
+            if let Some(source_id_value) = source_id {
+                if let Some(target_id_value) = target_id {
+                    graph.add_edge(*source_id_value, *target_id_value, ());
+                    edges_by_node_pair.insert((*source_id_value, *target_id_value), edge.clone());
+                }
+            }
+        }
+
+        (graph, edges_by_node_pair)
```
dkhofer, line 455: I don't love having to return edges_by_node_pair here, but I couldn't think of a more elegant way to get a mapping from one representation of edge (pair of black IDs) to the regular edge object

src/models/block_group.rs
```diff
@@ -255,218 +239,11 @@ impl BlockGroup {
         }
     }
 
-    pub fn get_block_boundaries(
```
dkhofer, line 258: Most of this is just refactoring, moving it into the Edge model, which I think is a better place for it.

src/models/collection.rs
```diff
@@ -0,0 +1,70 @@
+use rusqlite::types::Value;
```
dkhofer, line 1: Felt like putting Collection into its own file while I was at it

src/exports/gfa.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let tmp_dir = tempdir().expect("Couldn't get handle to temp directory");
+        let gfa_path = tmp_dir
+            .path()
+            .join("intermediate.gfa")
+            .into_os_string()
+            .into_string()
+            .unwrap();
+
+        export_gfa(&conn, collection_name, &gfa_path);
```
Chris7, line 204: just a fyi, you can use pathbuf and avoid the string conversions

src/models/edge.rs
Chris7: I think we can assign more information to edges than we have been, but we can do that later. the `()` part of the DiGraphMap is edge storage iirc. So we could put some custom structure there.

src/exports/gfa.rs
dkhofer: Oh yeah, that does simplify things a bit, done here and with gfa imports

src/models/edge.rs
dkhofer: Oh, good to know.  I'll keep that in mind

PR #43, opened 2024-09-16
Graph model docs - port graphs & segment graphs

bobvh:
Python notebook to help settle model questions. The port graph section describes the current model (I think), with the exception that I do give sequence nodes their own index to allow multiple nodes to reference the same hash. The segment graph section describes how to convert from the port graph format to something you'd see in a GFA. Each model has their own strengths (e.g. not needing node splitting in the port model, compatibility with graph aligners in the segment model). Let me know what you think of the visualizations, it's really graphviz doing the heavy lifting, so if needed we could have the Rust client output a dot file for graphviz too.


doc/gen graph model explained.ipynb
```diff
@@ -0,0 +1,688 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# The port graph model\n",
+    "\n",
+    "We use a graph data model where nodes represent sequence segments, and edges indicate how multiple segments can be joined together. More specifically, we use a \"port graph\" in which each node is accessed via ports that correspond to nucleotide positions on the source sequence. This allows pairs of edges to reference subsequences without having to split up existing nodes. The port information is stored on the edges, which specify 'to' and 'from' positions in addition to the source and target node.\n",
+    "\n",
+    "The actual sequences are stored in a hash table that may contain metadata on where to find the source sequence (drive, url, ...), but lacks any biological context. Sequence nodes are instantiations of source sequences, they do not have to be unique and are well suited to carry metadata. They are referenced by an identifier that is unique within the graph, which avoids creating ambiguous edges or cycles."
```
dkhofer, line 22: In CS, "hash table" has a specific meaning that's not quite the same as how we store sequences.  I'd suggest stating it like "The actual sequences are stored with their hashes in a database table..."

dkhofer:
I think this is a decent first pass at describing the data model.  I think if I didn't know about gen but were knowledgeable about coding and bioinformatics, this would be helpful.  I think longer term it might be good to extract the diagrams and explanations from the python code and put them in a separate doc that isn't a notebook.  If I were trying to understand the data model I'd be wondering why it's being laid out in python when the gen code is in rust, although I understand it can help to make the details clearer when explaining the model.

doc/gen graph model explained.ipynb
```diff
@@ -0,0 +1,688 @@
...
+    "\n",
+    "draw_port_graph(G)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# The segment graph model\n",
+    "The port graph as described above is an efficient way to represent and combine many operations, but it is less intuitive compared to a pangenome graph where each node is a distinct segment. The latter form is also what graph aware sequence aligners tend to expect. That being said, we can convert a port graph into a segment graph and vice versa without loss of information."
```
dkhofer, line 436: I think you mentioned in the PR description that segment graphs are pretty much how GFAs are formatted.  It'd be good to mention that here to provide more context for the reader

dkhofer:
Instead of checking in a db file, could you maybe instead add the fasta or whatever files you used to create it into a subdirectory of `doc`, and then put instructions for running gen to recreate the db from the files?

doc/gen graph model explained.ipynb
bobvh: TIL hash tables use a hash _of_ the key and not _as_ the key. Fixed.

bobvh:
> Instead of checking in a db file, could you maybe instead add the fasta or whatever files you used to create it into a subdirectory of `doc`, and then put instructions for running gen to recreate the db from the files?
The DB file shouldn't have gotten added, I had my setup wrong.

bobvh:
@dkhofer I'll convert this to a regular markdown file and refer to saved SVG files instead of the notebook. On the rust vs python confusion: can I keep the python notebook that generated the SVGs on the repo, or keep that out too?

dkhofer:
I'm fine with keeping the python notebook.  Maybe add a comment at the top saying it's a python port of the rust models/logic?

PR #44, opened 2024-09-16
Separate operation table POC

Chris7:
This makes a separate database for handling operations done to a database for genetic changes. The BASE_DIR change is mostly here for testing so we can set a temp directory for tests to run and not have issues due to asset cleanup.

I don't really love the operation_connection thing and having 2 separate connections and needing to know which one to use. I did try briefly making the connections a global configuration, but after thinking about behavior during threading i think it's a bad idea.


src/operation_management.rs
```diff
@@ -72,8 +53,9 @@ pub fn apply_changeset(conn: &Connection, op_id: i32) {
     conn.pragma_update(None, "foreign_keys", "1").unwrap();
 }
 
-pub fn revert_changeset(conn: &Connection, op_id: i32) {
-    let change_path = get_changeset_path(conn).join(format!("{op_id}.cs"));
+pub fn revert_changeset(conn: &Connection, operation: &Operation) {
+    let change_path =
+        get_changeset_path(operation).join(format!("{op_id}.cs", op_id = operation.id));
```
dkhofer, line 58: What does the cs suffix stand for?  Changeset?

dkhofer (approved):

src/operation_management.rs
Chris7: exactly that!

PR #46, opened 2024-09-18
Readme intro: user interpretation of the data model and future features

bobvh:
This is how I'm viewing what users would use the different objects we expose for. I refer to "block groups" as "pangenomic molecules" to try to make it more clear that "chr1 and its derivatives" is a block group. For samples I went a different direction than what's currently in the schema.


anderson_out.gfa
```diff
@@ -0,0 +1,33 @@
+S	0	GG	*
```
dkhofer, line 1: It'd be good to not have this at the top level of the repo.  Actually, is this even something you want to be checking in?

README.md
```diff
@@ -1 +1,69 @@
-Experiment
+# Gen
```
dkhofer, line 1: So happy to finally get a real readme here!

README.md
```diff
@@ -1 +1,69 @@
...
+<!-- -Translating coordinates between paths -->
+
+### Recording sequence changes
+<!-- From a VCF file -->
+Sequence variants observed through NGS can be imported into a gen repository via standard VCF file obtained from variant callers like Freebayes, GATK, or DeepVariant. [...]
+<!-- -From a sequence file that was edited externally -->
+
+<!-- -From the gen command line -->
+
+Inserting one or more genetic parts at a specific locus can be done using the `gen update <location> <sequence>` command. The location parameter is given as the combination of a path (molecule) identifier and a numerical coordinate referenced to that path, separated by a colon. If the option `--propagate` is used, all other paths that include the location will be updated as well. By default the molecule itself is changed, but by using the option `--new` new molecules are created and their path IDs are returned to the user. Multiple inserts can be specified using the `--fasta <file>` argument. In this case new path IDs are always returned to the user, but unless the option `--new` is given the original path is removed.
```
dkhofer, line 51: I don't think we've implemented any of this yet.  Maybe you could take this part out for now and make sure there are tasks to implement all these options?

dkhofer:
Overall looks good, but had a few comments.  Thanks!

README.md
bobvh: Correct, I was trying to see if writing docs and tasks could be combined since we're pre-release, but it's not really the right place (especially with the part about path rewriting not being compatible with immutable paths). I'll comment it out and use the private task board.

anderson_out.gfa
bobvh: No this shouldn't be checked in, I misunderstood that VS Code did a `git add` automatically (or misunderstood git). Will go through them with more git rm in next commit.

PR #47, opened 2024-09-18
Apply command

Chris7:
Setup for applying a changeset to a given branch.

### DIFF
--- a/src/models/operations.rs
+++ b/src/models/operations.rs
@@ -442,7 +442,6 @@ impl OperationState {
           UPDATE SET branch_id=excluded.branch_id;",
             )
             .unwrap();
-        println!("setting branc to {branch_name}");
         stmt.execute(params_from_iter(vec![
             Value::from(db_uuid.to_string()),
             Value::from(branch.id),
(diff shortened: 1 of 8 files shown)


src/models/operations.rs
```diff
@@ -342,13 +352,49 @@ impl Branch {
...
+        let branch = Branch::get_by_id(conn, branch_id)
+            .unwrap_or_else(|| panic!("No branch with id {branch_id}."));
+        let graph = Operation::get_operation_graph(conn);
+        let mut operations: Vec<Operation> = vec![];
+        let creation_id = branch.start_operation_id.unwrap_or(1);
+
+        let mut dfs = DfsPostOrder::new(&graph, creation_id);
+
+        // Traverse all transitive predecessors
+        while let Some(ancestor) = dfs.next(&graph) {
```
dkhofer, line 387: I'm confused, this is adding all operations to the return vec, and then below, we're adding just those operations for the input branch, which I assume would all be redundant

dkhofer: Never mind, just saw the creation_id thing

dkhofer (approved):
LGTM

PR #48, opened 2024-09-18
Gfa path export

dkhofer:


migrations/core/01-initial/up.sql
```diff
@@ -85,7 +85,7 @@ CREATE TABLE path_edges (
   FOREIGN KEY(edge_id) REFERENCES edges(id),
   FOREIGN KEY(path_id) REFERENCES path(id)
 ) STRICT;
-CREATE UNIQUE INDEX path_edges_uidx ON path_edges(path_id, edge_id);
+CREATE UNIQUE INDEX path_edges_uidx ON path_edges(path_id, edge_id, index_in_path);
```
dkhofer, line 88: This resolves the constraint violations that Bob saw

dkhofer: The same sequence (or pair of edges) can show up multiple times in a path with this

src/models/path.rs
```diff
@@ -119,6 +119,11 @@ impl Path {
         paths
     }
 
+    pub fn get_paths_for_collection(conn: &Connection, collection_name: &str) -> Vec<Path> {
+        let query = "SELECT path.id, path.block_group_id, path.name FROM path JOIN block_group ON path.block_group_id = block_group.id WHERE block_group.collection_name = ?1";
```
Chris7, line 123: i geneally always do `select *....` since the column order will then match the table order and we don't get bad unpacking of queries if/when the models change.

src/models/path_edge.rs
```diff
@@ -92,6 +94,48 @@ impl PathEdge {
             .map(|edge_id| edges_by_id[&edge_id].clone())
             .collect::<Vec<Edge>>()
     }
+
+    pub fn edges_for_paths(conn: &Connection, path_ids: Vec<i32>) -> HashMap<i32, Vec<Edge>> {
+        let placeholder_string = path_ids.iter().map(|_| "?").join(",");
+        let path_edges = PathEdge::query(
+            conn,
+            format!(
+                "select id, path_id, index_in_path, edge_id from path_edges where path_id in ({}) ORDER BY path_id, index_in_path",
```
Chris7, line 103: same thing wrt unpacking columns as above.

Chris7 (approved):
it looks good. Only comment is about a convention i've been trying to use for column table order in our generic query methods

src/models/path_edge.rs
dkhofer: Changed both

PR #50, opened 2024-09-20
Custom changeset management

Chris7:
This is an extension to the changeset management to identify dependencies of a changeset and write those out. On imports, we ensure that the ids of dependent entities are created or retrieved from the current checked out database.

The workflow this handles is suppose we have branch A with operations 1->2->3, and branch B with operations 1->4->5. If we apply operation 3, things required from operation 2 may not be present. This ensures they will be. For the reversion, we create a new commit with the updates from this application (similar to a new commit id resulting from a cherry-pick).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,6 @@ noodles = { version = "0.78.0", features = ["core", "vcf", "fasta", "async"] }
 petgraph = "0.6.5"
 chrono = "0.4.38"
 tempdir = "0.3.7"
+fallible-streaming-iterator = "0.1.9"
+serde = {  version = "1.0.210", features = ["derive"] }
+serde_json = "1.0.128"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,7 +49,7 @@ pub fn parse_genotype(gt: &str) -> Vec<Option<Genotype>> {
         true => Phasing::Unphased,
         false => Phasing::Phased,
     };
-    for entry in gt.split_inclusive(|c| c == '|' || c == '/') {
+    for entry in gt.split_inclusive(['|', '/']) {
         let allele;
         let mut phasing = Phasing::Unphased;
         if entry.ends_with(['/', '|']) {
--- a/src/models/block_group.rs
+++ b/src/models/block_group.rs
@@ -1,7 +1,9 @@
+use std::collections::{HashMap, HashSet};
+
 use intervaltree::IntervalTree;
 use petgraph::Direction;
 use rusqlite::{params_from_iter, types::Value as SQLValue, Connection};
-use std::collections::{HashMap, HashSet};
+use serde::{Deserialize, Serialize};
 
 use crate::graph::all_simple_paths;
 use crate::models::block_group_edge::BlockGroupEdge;
@@ -11,7 +13,7 @@ use crate::models::path_edge::PathEdge;
 use crate::models::sequence::Sequence;
 use crate::models::strand::Strand;
 
-#[derive(Debug)]
+#[derive(Debug, Deserialize, Serialize)]
 pub struct BlockGroup {
     pub id: i32,
     pub collection_name: String,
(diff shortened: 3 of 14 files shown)


Cargo.toml
```diff
@@ -17,3 +17,6 @@ noodles = { version = "0.78.0", features = ["core", "vcf", "fasta", "async"] }
 petgraph = "0.6.5"
 chrono = "0.4.38"
 tempdir = "0.3.7"
+fallible-streaming-iterator = "0.1.9"
+serde = {  version = "1.0.210", features = ["derive"] }
+serde_json = "1.0.128"
```
Chris7, line 22: we can pick a more efficient format once we settle on this.

src/models/path.rs
```diff
@@ -1,15 +1,18 @@
...
 use rusqlite::types::Value;
 use rusqlite::{params_from_iter, Connection};
-use std::collections::{HashMap, HashSet};
+use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+use crate::models::{edge::Edge, path_edge::PathEdge, sequence::Sequence, strand::Strand, Sample};
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Deserialize, Serialize)]
 pub struct Path {
```
Chris7, line 12: reorganized so it matches the order in up.sql

src/models/path.rs
```diff
@@ -76,7 +81,33 @@ impl Path {
                 })
             })
             .unwrap();
-        let path = rows.next().unwrap().unwrap();
+        let path = match rows.next().unwrap() {
```
Chris7, line 84: made creation get-or-create

src/models/block_group.rs
```diff
@@ -105,17 +106,29 @@ impl BlockGroup {
             Err(rusqlite::Error::SqliteFailure(err, details)) => {
                 if err.code == rusqlite::ErrorCode::ConstraintViolation {
                     println!("{err:?} {details:?}");
-                    BlockGroup {
-                        id: conn
+                    let bg_id = match(sample_name) {
```
Chris7, line 109: this was bugged behavior if the insert was conflicting it would always return the null sample

src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub fn parse_genotype(gt: &str) -> Vec<Option<Genotype>> {
         true => Phasing::Unphased,
         false => Phasing::Phased,
     };
-    for entry in gt.split_inclusive(|c| c == '|' || c == '/') {
+    for entry in gt.split_inclusive(['|', '/']) {
```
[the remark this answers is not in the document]
dkhofer: Time to revolt

src/models/edge.rs
```diff
@@ -528,6 +545,79 @@ mod tests {
         assert_eq!(edge_result3.target_coordinate, -1);
     }
 
+    #[test]
+    fn test_bulk_create_returns_edges_in_order() {
```
dkhofer, line 549: Oh, was bulk create not returning them in order?  Oops, thanks!

src/models/path.rs
```diff
@@ -65,8 +68,10 @@ pub struct NewBlock {
 
 impl Path {
     pub fn create(conn: &Connection, name: &str, block_group_id: i32, edge_ids: &[i32]) -> Path {
+        // TODO: Should we do something if edge_ids don't match here?
```
dkhofer, line 71: I don't understand what is meant by "if edge_ids don't match", can you add a bit more?  I have a list of a few validations I'd like to add, I'll bring them up on slack

src/models/path.rs
dkhofer: :+1:

src/models/sequence.rs
```diff
@@ -30,6 +31,17 @@ pub struct NewSequence<'a> {
     shallow: bool,
 }
 
+impl<'a> From<&'a Sequence> for NewSequence<'a> {
```
dkhofer, line 34: Thanks for this too

src/operation_management.rs
```diff
@@ -31,28 +54,356 @@ pub fn get_file(path: &PathBuf, mode: FileMode) -> fs::File {
...
+                    let source_hash =
+                        str::from_utf8(item.new_value(1).unwrap().as_bytes().unwrap())
+                            .unwrap()
+                            .to_string();
+                    let target_hash =
+                        str::from_utf8(item.new_value(4).unwrap().as_bytes().unwrap())
+                            .unwrap()
+                            .to_string();
+                    created_edges.insert(edge_pk);
+                    if !created_sequences.contains(&source_hash) {
```
dkhofer, line 112: How are we sure that this sequence doesn't show up later as a created sequence?

Cargo.toml
dkhofer: :+1:

dkhofer (approved):
This is awesome.  LGTM, no blocking comments

PR #54, opened 2024-09-25
Topological correctness

dkhofer:


fixtures/aa.gfa
```diff
@@ -0,0 +1,5 @@
+H	VN:Z:1.2
+S	1	A	SN:Z:123	SO:i:0	SR:i:0
+S	2	A	SN:Z:123	SO:i:0	SR:i:0
+L	1	+	2	+	*
+P	124	1+,2+	4M
```
Chris7, line 5: i'm not sure what the parser would do, but this should be 1M (1 match)

bobvh: Why? Shouldn't it be 0M for overlap of length 0 if it's representing the sequence AA?

Chris7: yeah, it should be. I knew 4m was way too much :)

src/imports/fasta.rs
```diff
@@ -57,13 +64,16 @@ pub fn import_fasta(
                     .sequence(&sequence)
                     .save(conn)
             };
+            let node = Node::create(conn, &seq.hash);
```
Chris7, line 67: So if i import the same file 2x, we'll have duplicate nodes + edges made?

src/models/node.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl Node {
+    pub fn create(conn: &Connection, sequence_hash: &str) -> Node {
+        let insert_statement = format!(
+            "INSERT INTO nodes (sequence_hash) VALUES ('{}');",
+            sequence_hash
+        );
+        let _ = conn.execute(&insert_statement, ());
+        Node {
+            id: conn.last_insert_rowid() as i32,
```
Chris7, line 26: is there a reason to do this over returning id?

src/imports/fasta.rs
dkhofer: Yeah.  It's not great, but I don't see it as blocking.  I think properly detecting the duplicates in general would take some work, worth doing long term.

fixtures/aa.gfa
dkhofer: Fixed

src/models/node.rs
dkhofer: I genuinely thought I was using the full node somewhere, but I guess not.  Changed

src/imports/fasta.rs
bobvh: Is it that much of a problem on the fasta import side? When would people import the same fasta file and expect them to collapse onto the same nodes and edges? I can understand on the vcf import you'd eventually want a heuristic that says something like "for a pair of edges (p, q), if there is already a pair of edges (a, b) with a shared node id (from the reference path) and identical coordinates on all side, then check to see if the new node has the same hash as the old node."

migrations/core/01-initial/up.sql
```diff
@@ -88,19 +94,22 @@ CREATE TABLE path_edges (
...
     hex(randomblob(4)) || '-' || hex(randomblob(2)) || '-' || '4' ||
     substr(hex( randomblob(2)), 2) || '-' ||
     substr('AB89', 1 + (abs(random()) % 4) , 1)  ||
     substr(hex(randomblob(2)), 2) || '-' ||
     hex(randomblob(6))
   ));
+INSERT INTO sequence (hash, sequence_type, sequence, name, file_path, "length") values ("start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "OTHER", "start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "", "", 64), ("end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "OTHER", "end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "", "", 64);
+INSERT INTO nodes (id, sequence_hash) values (1, "start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy");
+INSERT INTO nodes (id, sequence_hash) values (2, "end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz");
+UPDATE SQLITE_SEQUENCE SET seq = 2 WHERE name = 'nodes';
```
Chris7, line 115: is this still needed?

src/models/node.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl Node {
+    pub fn create(conn: &Connection, sequence_hash: &str) -> i32 {
+        let insert_statement = format!(
+            "INSERT INTO nodes (sequence_hash) VALUES ('{}');",
+            sequence_hash
+        );
+        let _ = conn.execute(&insert_statement, ());
+        conn.last_insert_rowid() as i32
```
Chris7, line 23: i have a bad feeling about using this instead of returning id. It seems like we're opening ourselves up to race conditions around this connection session and insertion execution order.

src/imports/gfa.rs
```diff
@@ -323,9 +344,115 @@ mod tests {
...
+        let part15_16 = vec!["A", "T"];
+        let part17 = "GG";
+        let part18_19 = vec!["T", "G"];
+        let part20 = "A";
+        let part21_22 = vec!["T", "C"];
+        let part23_24 = vec!["A", "T"];
+        let part25_26 = vec!["A", "G"];
+
+        let mut expected_sequences = HashSet::new();
+        for part_a in &part4_5 {
```
Chris7, line 389: i wonder if this can be all done w/ combinations/chain. No need to, just a thought.

Chris7 (approved):
Biggest concern is the last_id thing biting us in the ass if we add concurrency eventually. I don't feel good about getting the id by a is-probably-right mechanism vs. one we know to be.

Eventually, we'll need to see how to make node creation stateless wrt uniqueness. I think that using from_node_id + alt seq + to_node_id may accomplish that. Thought is that we should expect multiple vcfs to be loaded over time and we should expect overlap between each vcf file. We shouldn't be duplicating those entries every time. I think that would also dedupe fastas as well (hash on from node (source), seq, to node (sink) and name i think may get it done, maybe need collection too.).

src/imports/fasta.rs
dkhofer: Probably not an issue for fasta, but definitely VCF and GFA, since GFA doesn't do anything to canonically label segments and links

src/models/node.rs
dkhofer: My bad, I let autopilot write a lot of this code and I zoned on this.  Change to use `returning`

migrations/core/01-initial/up.sql
dkhofer: I guess not, so that's nice.  Removed

dkhofer:
Definitely agree that we need the deduping.  I'm thinking about it more and am open to your suggestion about from_node_id + alt seq + to_node_id, but I need to convince myself that that's all that's needed.  I don't have a specific example in mind but I feel like duplicates could pop up more generally, and maybe we need something stronger to handle more general cases.

PR #55, opened 2024-09-25
Branch bifurcation protection

Chris7:
Prevents users adding operations in the middle of a branch.

### DIFF
--- a/src/models/operations.rs
+++ b/src/models/operations.rs
@@ -5,6 +5,7 @@ use petgraph::visit::DfsPostOrder;
 use petgraph::Direction;
 use rusqlite::types::Value;
 use rusqlite::{params_from_iter, Connection};
+use std::string::ToString;
 
 #[derive(Clone, Debug)]
 pub struct Operation {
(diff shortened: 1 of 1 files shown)


src/models/operations.rs
```diff
@@ -653,4 +672,123 @@ mod tests {
             .collect::<Vec<i32>>();
         assert_eq!(ops, vec![1, 6, 7, 8, 9, 10, 11]);
     }
+
+    #[test]
+    #[should_panic(
+        expected = "The current operation is in the middle of a branch. A new operation would create a bifurcation in the branch lineage. Create a new branch if you wish to bifurcate the current set of operations."
```
dkhofer, line 678: Maybe make this string a constant?

dkhofer (approved):
LGTM, comment is non-blocking

src/models/operations.rs
Chris7: i tried https://internals.rust-lang.org/t/passing-variables-or-constants-as-arguments-to-the-should-panic-expected-attribute-macro/16695 :(

PR #58, opened 2024-09-30
Cache fasta index for external sequence lookup

Chris7:
Caches the index file so it doesn't need to be read everytime for external sequence lookup.


dkhofer (approved):
Cool.

PR #59, opened 2024-10-01
Use 64 bits

dkhofer:


dkhofer:
This is just a straight-up global replacement of i32/u32 with i64/u64.  I had to fix a few minor clippy errors in operation management but that's it

Chris7 (approved):
We should follow up to refine this by trimming down some fields to what they actually need (like allele doesn't need a i64 and most should be u64 too for ids)

PR #63, opened 2024-10-02
Speed up GFA import, add chr22.gfa benchmark file, fix GFA export formatting issues

dkhofer:


src/exports/gfa.rs
```diff
@@ -181,9 +185,14 @@ fn write_paths(
             }
         }
 
+        let full_path_name = if sample_name.is_some() {
```
Chris7, line 188: i'm kind of surprised the linter let you do this, it always yells at me to do `if let Some(full_path_name) = sample_name``

src/exports/gfa.rs
```diff
@@ -181,9 +185,14 @@ fn write_paths(
...
+            format!("{}.{}", path.name, sample_name.unwrap()).to_string()
```
Chris7, line 189: can we check if sample_name is actually the empty string so we don't prefix the reference chromosomes w/ `.`

Chris7 (approved):
LGTM. One issue on the path name convention to handle the null sample case.

Chris7:
oh, can you sneak the strand::forwrd change in here too for boundary edges?

dkhofer:
> oh, can you sneak the strand::forwrd change in here too for boundary edges?

Yep, snuck

PR #64, opened 2024-10-04
GFA export improvements

dkhofer:
There's about 4 things happening here:
1. Changing segment ID to be <node ID>.<sequence start coordinate>
2. Copy/pasting in the GFA reader code from the third party crate (MIT license) so I can fix a bug with parsing a string segment ID
3. Adding a --sample flag to GFA export to limit to just one sample
4. Replacing some manual code for filtering out start/end nodes with a dedicated method

### DIFF
--- a/src/exports/gfa.rs
+++ b/src/exports/gfa.rs
@@ -1,6 +1,6 @@
 use itertools::Itertools;
 use petgraph::prelude::DiGraphMap;
-use rusqlite::Connection;
+use rusqlite::{types::Value as SQLValue, Connection};
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::io::{BufWriter, Write};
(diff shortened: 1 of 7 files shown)


src/exports/gfa.rs
```diff
@@ -11,19 +11,49 @@ use crate::models::{
...
-    for block_group in block_groups {
-        let block_group_edges = BlockGroupEdge::edges_for_block_group(conn, block_group.id);
-        edge_set.extend(block_group_edges.into_iter());
+    if sample_name.is_some() {
+        let block_groups = BlockGroup::query(
+            conn,
+            "select * from block_group where collection_name = ?1 AND sample_name = ?2;",
+            vec![
+                SQLValue::from(collection_name.to_string()),
+                SQLValue::from(sample_name.clone().unwrap()),
```
Chris7, line 38: i think if you change this to `if let Some(name) = sample` above you can just use that value w/o the clone

src/exports/gfa.rs
```diff
@@ -73,24 +96,36 @@ fn write_segments(
...
+    // NOTE: We encode the node ID and start coordinate in the segment ID
+    format!("S\t{}.{}\t{}\t*\n", node_id, sequence_start, sequence,)
 }
 
 fn write_links(
     writer: &mut BufWriter<File>,
     graph: &DiGraphMap<i64, ()>,
     edges_by_node_pair: &HashMap<(i64, i64), Edge>,
-    terminal_block_ids: &HashSet<i64>,
+    start_segments_by_end_segment: HashMap<(i64, i64), i64>,
```
Chris7, line 108: i don't really get this name. Can you throw a comment above it? I think that will start filling in our docs too

src/gfa_reader.rs
```diff
@@ -0,0 +1,701 @@
+use std::fs::File;
```
Chris7, line 1: i think we need to keep the attribution/link to the source

Chris7 (approved):
LGTM. minor stuff

bobvh:
I confirmed this also solves the orphan nodes bug in the aleDB non-standard VCF dataset.

src/exports/gfa.rs
dkhofer: Good call, done

src/exports/gfa.rs
dkhofer: I renamed it, and added a comment in the code below.  It's a minor complication from how I'm defining segment ID as (node ID + sequence start coordinate).  An edge's source is (node ID + sequence end coordinate), so we need to do a little more work to convert that to the source segment ID.  The target segment ID for an edge is already (node ID + sequence start coordinate) so it's fine

src/gfa_reader.rs
dkhofer: I've never done that before, so I just copied the full text from https://github.com/MoinSebi/gfa-reader?tab=MIT-1-ov-file#readme , hope that works

PR #65, opened 2024-10-07
Handle cases where branch on secondary db showed commit 1

Chris7:
If we were working off a second db, operation id 1 would always show up even if it was not part of that db.

### DIFF
--- a/src/models/operations.rs
+++ b/src/models/operations.rs
@@ -8,7 +8,7 @@ use rusqlite::{params_from_iter, Connection};
 use std::collections::HashSet;
 use std::string::ToString;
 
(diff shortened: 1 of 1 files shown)


src/models/operations.rs
```diff
@@ -8,7 +8,7 @@ use rusqlite::{params_from_iter, Connection};
 use std::collections::HashSet;
 use std::string::ToString;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
dkhofer, line 11: Maybe also add Eq and Hash here?

dkhofer (approved):
LGTM, only one minor comment

PR #67, opened 2024-10-08
Node dedupe

Chris7:


src/models/edge.rs
```diff
@@ -113,7 +113,7 @@ impl Edge {
         phased: i64,
     ) -> Edge {
         let query = "INSERT INTO edges (source_node_id, source_coordinate, source_strand, target_node_id, target_coordinate, target_strand, chromosome_index, phased) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8) RETURNING *";
-        let id_query = "select id from edges where and source_node_id = ?1 and source_coordinate = ?2 and source_strand = ?3 and target_node_id = ?4 and target_coordinate = ?5 and target_strand = ?6 and chromosome_index = ?7 and phased = ?8";
```
dkhofer, line 116: Oof, can't believe that was valid

src/updates/vcf.rs
```diff
@@ -287,7 +287,16 @@ pub fn update_with_vcf(
             let sequence =
                 SequenceCache::lookup(&mut sequence_cache, "DNA", vcf_entry.alt_seq.to_string());
             let sequence_string = sequence.get_sequence(None, None);
-            let node_id = Node::create(conn, sequence.hash.as_str());
+            let node_id = Node::create(
+                conn,
+                sequence.hash.as_str(),
+                calculate_hash(&format!(
+                    "{bg_id}.{path_id}:{ref_start}-{ref_end}->{sequence_hash}",
```
dkhofer, line 294: I need to ask the stupid but obvious question: Are we sure that caching this information as a hash is better performance-wise than not adding the hash and just querying for the nodes directly?  Is that what you and Bob were discussing trying on slack earlier?

dkhofer (approved):
LGTM, approving, but it'd be good to test this against whatever Bob provides that reproduces the problem for him

src/updates/vcf.rs
dkhofer: :+1:

Cargo.toml
```diff
@@ -23,3 +26,4 @@ serde_json = "1.0.128"
 sha2 = "0.10.8"
 tempfile = "3.12.0"
 tempdir = "0.3.7"
+cargo-llvm-cov = "0.6.13"
```
dkhofer, line 29: Would you mind adding a coverage job here?  https://github.com/ginkgobioworks/gen/blob/main/.github/workflows/test.yml

src/imports/fasta.rs
```diff
@@ -176,4 +184,36 @@ mod tests {
             "ATCGATCGATCGATCGATCGGGAACACACAGAGA".to_string()
         );
     }
+
+    #[test]
+    fn test_deduplicates_nodes() {
+        setup_gen_dir();
+        let mut vcf_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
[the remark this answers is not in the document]
dkhofer: I might be missing something, but it still seems like this isn't needed?

dkhofer:
Still LGTM.  Thanks for making the benchmarking tests opt-in, the one that was failing a lot was bugging me

bobvh:
The block_group id:path id pair doesn't have a 1:1 mapping to the reference sequence, since during the update new paths are created.

Minimal example included here: https://github.com/ginkgobioworks/gen/commit/ab68650f963abf88b9e98652c0a0f2ec2f5f7484

It would need to be deduped on a path in the sample-less block_groups.

PR #68, opened 2024-10-08
Add in combinatorial code

dkhofer:


src/models/edge.rs
```diff
@@ -483,6 +467,30 @@ impl Edge {
 
         (graph, edges_by_node_pair)
     }
+
+    pub fn boundary_edges_from_sequences(
```
dkhofer, line 471: I didn't like returning two separate things from the above method, felt like splitting out the boundary edge computation into a separate method

src/updates/library.rs
```diff
@@ -0,0 +1,243 @@
...
+        operation_conn,
+        &db_uuid,
+        name.to_string(),
+        "csv_update",
+        change.id,
+    );
+
+    let mut parts_reader = fasta::io::reader::Builder.build_from_path(parts_file_path)?;
+
+    let path = Path::get_paths(
```
dkhofer, line 47: Currently assumes there is only one path with the input name in the db, need to think about how to better handle this.  I don't want users to have to specify a block group ID

src/updates/library.rs
```diff
@@ -0,0 +1,243 @@
...
+            target_node_id: end_block.node_id,
+            target_coordinate: node_end_coordinate,
+            target_strand: Strand::Forward,
+            chromosome_index: 0,
+            phased: 0,
+        };
+        new_edges.insert(edge);
+    }
+
+    for part1 in &parts1 {
```
dkhofer, line 148: Creates all the edges for the combinatorial design unconditionally.  Can add in a flag to just add the specified part orderings if Bob wants

src/updates/library.rs
```diff
@@ -0,0 +1,243 @@
...
+            };
+            new_edges.insert(edge);
+        }
+    }
+
+    let new_edge_ids = Edge::bulk_create(conn, new_edges.iter().cloned().collect());
+    BlockGroupEdge::bulk_create(conn, path.block_group_id, &new_edge_ids);
+
+    let path_changes_count = parts1.len() * parts2.len();
+    let summary_str = format!("{path_name}: {path_changes_count} changes.\n");
```
dkhofer, line 169: Not sure exactly what to put here, I'm open to suggestions

src/main.rs
```diff
@@ -60,13 +61,28 @@ enum Commands {
...
+        library: Option<String>,
+        /// A fasta with the combinatorial library parts
+        #[arg(short, long)]
+        parts: Option<String>,
+        /// The name of the path to add the library to
+        #[arg(short, long)]
+        path_name: Option<String>,
+        /// The start coordinate for the region to add the library to
+        #[arg(short, long)]
+        start: Option<i64>,
```
bobvh, line 82: VCF imports fail now on a collision between -s for sample and -s for start

`    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/gen update --vcf fixtures/simple.vcf`
thread 'main' panicked at /Users/bvh/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.5.17/src/builder/debug_asserts.rs:112:17:
Command update: Short option names must be unique for each argument, but '-s' is in use by both 'sample' and 'start'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

src/updates/library.rs
Chris7: could do a blockgroup name + path instead of id. BG would be something like "X-vector", path would be backbone or something

src/main.rs
dkhofer: Should be fixed now, I just removed the short option for start and also for parts (which collided with path-name).  So you need to specify --start and --parts instead of -s and -p, but hopefully that isn't a big deal

src/exports/gfa.rs
```diff
@@ -56,8 +56,19 @@ pub fn export_gfa(
...
-    let (blocks, boundary_edges) = Edge::blocks_from_edges(conn, &edges);
 
+    let blocks = Edge::blocks_from_edges(conn, &edges);
+    let node_blocks_by_id: HashMap<i64, Vec<GroupBlock>> =
+        blocks
+            .clone()
+            .into_iter()
+            .fold(HashMap::new(), |mut acc, block| {
+                acc.entry(block.node_id)
+                    .and_modify(|blocks| blocks.push(block.clone()))
```
Chris7, line 67: if we're cloning the blocks here too, is there a reason to clone the blocks above? It seems like we should be able to just clone the pointer once here vs. cloning the vector then cloning the cloned block.

src/graph.rs
```diff
@@ -40,3 +40,125 @@ where
         None
     })
 }
+
+#[cfg(test)]
+mod tests {
```
Chris7, line 45: very glad these all worked!

src/main.rs
```diff
@@ -60,13 +61,28 @@ enum Commands {
...
         /// If no sample is provided, enter the sample to associate variants to
         #[arg(short, long)]
         sample: Option<String>,
+        /// A CSV with combinatorial library information
+        #[arg(short, long)]
+        library: Option<String>,
+        /// A fasta with the combinatorial library parts
+        #[arg(long)]
+        parts: Option<String>,
+        /// The name of the path to add the library to
```
Chris7, line 77: you might be able to reuse noodle's region parser to combine all this so users can specify it like
```
chr1:123-456
```
or ```chr1:245`

src/updates/library.rs
```diff
@@ -0,0 +1,259 @@
...
+    operation_management::attach_session(&mut session);
+    let change = FileAddition::create(operation_conn, library_file_path, FileTypes::CSV);
+
+    let db_uuid = metadata::get_db_uuid(conn);
+
+    let operation = Operation::create(
+        operation_conn,
+        &db_uuid,
+        name.to_string(),
+        "csv_update",
```
Chris7, line 42: we might want to be more specific here for the operation -- like `library_csv_update` if we have lots of routes giving csvs.

Chris7 (approved):
LGTM! just a few comments, nothing blocking.

src/main.rs
dkhofer: @bobvh what do you think of Chris's suggestion?  It would basically remove the start and end flags, and I think the path name flag.

However, the parts fasta file has multiple entries, I'm not sure there's an obvious one to put that information into.  The entries in the parts file don't correspond directly to the contig they're being added to

src/exports/gfa.rs
dkhofer: I realized that I copy and pasted this block of code in a few places, and it's being used only before calls to the `boundary_edges_from_sequences` method, so I just refactored it there, and boom, all the needs to clone disappeared.

src/main.rs
```diff
@@ -214,22 +230,40 @@ fn main() {
...
+                    &conn,
+                    &operation_conn,
+                    name,
+                    &path_name.clone().unwrap(),
+                    start.unwrap(),
+                    end.unwrap(),
+                    &parts.clone().unwrap(),
+                    library_path,
+                );
+            } else {
```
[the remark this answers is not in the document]
dkhofer: Done

src/updates/library.rs
dkhofer: Done

src/main.rs
dkhofer: Basically, I'm not seeing a good way to implement this.  I'm going to merge this PR so I can move on to other stuff, but if Bob has ideas I'll address them in a follow up PR

PR #69, opened 2024-10-11
Fix boundary edge bug

dkhofer:


src/models/edge.rs
```diff
@@ -330,14 +330,14 @@ impl Edge {
                 edges_by_source_node_id
                     .entry(edge.source_node_id)
                     .and_modify(|edges| edges.push(edge))
-                    .or_default();
+                    .or_insert(vec![edge]);
```
dkhofer, line 333: This and the below line are the source of the bug I found.  For the first time access of a given key, we were setting the value to empty vec instead of vec containing the edge. :-/

Chris7 (approved):
lgtm

PR #70, opened 2024-10-11
Yeast analysis example

bobvh:
The bcftools wrangling that was required to get the variant calls to end up in the right block groups makes me think that gen should support some of these reassigments. Contig names I can see as being out of scope, but for samples we should be able to either import a vcf into a sample regardless of the sample name in the vcf, or have an operation where a new block_group is created by taking the union of two block_groups.

This document should be expanded with more interpretation and downstream analysis of the GFA files, and use of gen branching and checkout when switching strategies from variant stacking to genome alignment.


examples/yeast_crosses/Analysis.md
```diff
@@ -0,0 +1,154 @@
...
+demonstrate a workflow that makes us of variant call files, and a workflow that uses whole genome alignment. In both
+cases we study two beer yeasts from the [1002 Yeast Genome collection](http://1002genomes.u-strasbg.fr/): Orval 
+trappist yeast strain DBVPG6695 from Belgium and American ale strain 1.3_Safale_US05 (codenames BRQ and CFD).
+
+First we set up a new gen repository to store and track our sequences. For this project we don't anticipate using
+more than one gen database file, and also don't need more than one collection. By setting defaults we won't have to
+enter these details in future commands.
+
+```console
+bvh@mbp:~$ gen init
```
Chris7, line 13: we should remove the prefixes from your cpu so all this code is copy/pasteable without additional work

Chris7 (approved):
i really like this.

PR #71, opened 2024-10-15
WIP: Lineage poc

Chris7:


src/models/block_group.rs
```diff
@@ -270,6 +336,14 @@ impl BlockGroup {
         }
     }
 
+    pub fn get_graph(conn: &Connection, block_group_id: i64) -> DiGraphMap<i64, ()> {
```
Chris7, line 339: this is a leftover from the little diff idea i was going to explore, but seems useful enough to keep around.

PR #72, opened 2024-10-16
Accession support

Chris7:
This creates a new model type, accession, that lets users assign a name to a subsequence of a path. There's support in here for a convention to use in vcf files for specifying that a change should be given a particular accession.


fixtures/simple.gfa
```diff
@@ -1,9 +1,9 @@
 H	VN:Z:1.2
```
Chris7, line 1: i changed this file to be the same sequence as simple.fa so we can use the same vcfs for both.

dkhofer: :+1: Cool

migrations/core/01-initial/up.sql
```diff
@@ -46,6 +46,43 @@ CREATE TABLE path (
 ) STRICT;
 CREATE UNIQUE INDEX path_uidx ON path(block_group_id, name);
 
+CREATE TABLE accession (
+  id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+  name TEXT NOT NULL,
+--  path accessions can reference other path accessions
+  path_id INTEGER NOT NULL,
+  accession_id INTEGER,
```
dkhofer, line 54: Is there some way to name this to make it clearer why we're referencing another accession?  parent_accession_id?  previous_accession_id?

src/models/accessions.rs
```diff
@@ -0,0 +1,347 @@
...
+                }
+            }
+            Err(_) => {
+                panic!("something bad happened.")
+            }
+        }
+    }
+}
+
+impl Query for Accession {
```
dkhofer, line 131: Nice

src/updates/vcf.rs
```diff
@@ -206,51 +212,92 @@ pub fn update_with_vcf(
...
-            let sample_bg_id = BlockGroupCache::lookup(
+            let mut sample_bg_id = BlockGroupCache::lookup(
                 &mut block_group_cache,
                 collection_name,
                 &fixed_sample,
                 seq_name.clone(),
             );
-            let sample_path = PathCache::lookup(&mut path_cache, sample_bg_id, seq_name.clone());
+            if sample_bg_id.is_err() {
+                println!("can't find sample bg....check this out more");
```
dkhofer, line 243: Maybe this should be a panic?

PR #73, opened 2024-10-16
Generate mappings of common ranges from one path to another

dkhofer:


src/models/path.rs
```diff
@@ -72,6 +74,13 @@ pub struct PathBlock {
     pub strand: Strand,
 }
 
+#[derive(Clone, Debug)]
+pub struct Annotation {
```
dkhofer, line 78: This is just to provide something to use in the annotation propagation.  When we implement GFF import/export we can use whatever feature struct is provided by noodles

src/range.rs
```diff
@@ -0,0 +1,301 @@
...
+        let overlaps = Range::find_pairwise_overlaps(self_intervals, other_intervals);
+
+        if overlaps.len() > 1 {
+            Range::consolidate_overlaps_about_the_origin(overlaps)
+        } else {
+            overlaps
+        }
+    }
+
+    fn find_pairwise_overlaps(intervals1: Vec<Range>, intervals2: Vec<Range>) -> Vec<Range> {
```
Chris7, line 126: i think if we sort these we could get a more effective overlap lookup by starting/stopped at the first/last indices of interval2

Chris7: should we merge some intervals based on some criteria? e.g. if we have [1,2], [2,3], [3,4], should it be [1,4]

Chris7: oh this is always going to be a list of 1/2 -- right

src/models/path.rs
```diff
@@ -283,14 +291,216 @@ impl Path {
...
+    pub fn intervaltree(&self, conn: &Connection) -> IntervalTree<i64, PathBlock> {
+        let blocks = self.blocks(conn);
         let tree: IntervalTree<i64, PathBlock> = blocks
             .into_iter()
             .map(|block| (block.path_start..block.path_end, block))
             .collect();
         tree
     }
+
+    pub fn find_block_mappings(&self, conn: &Connection, other_path: &Path) -> Vec<RangeMapping> {
```
Chris7, line 303: i think we need some kind of sequence extension logic option here. Imagine we have 2 gfas of identical sequences but different node lengths -- what happens here? I think the default should be everything is in the same frame/etc. but we should have a flag that is sequence based

Chris7: i wonder if we can actually resolve this with contig names and not need to ever fetch sequences here too

src/models/path.rs
```diff
@@ -283,14 +291,216 @@ impl Path {
...
+                                },
+                                target_range: Range {
+                                    start: their_start,
+                                    end: their_end,
+                                },
+                            };
+                            mappings.push(mapping);
+                        }
+
+                        if their_block.sequence_end < our_block.sequence_end {
```
Chris7, line 409: would we ever have a wrap around case that inadvertently hits the break condition?

src/models/path.rs
dkhofer: I think if we import from one GFA, then update from a second, we should generate the sequences from the paths defined in the second and check if there's an existing path that matches any of them, and if so insert appropriate edges etc. so that there's only one path through the block group that generates that sequence.

Is that the scenario you have in mind?

I think you're asking about an issue I brought up in my doc about reference paths: If a GFA is meant to update an existing path in the gen db, how does the user specify that.  The naive way would be to force them to redefine the entire path node-for-node as it exists in the database in their input GFA, and then they can define new segments etc off that in their input file.  They could of course export what's in the database and then add to the exported file, then update gen from that.

I do think it'll be possible to define an abbreviated format eventually, where they just specify new segments etc and specify where those would go on an existing path.

Regardless, I think your suggestion is out of scope for this PR?  This method is only meant to be used for paths that already exist in the database.  Resolving identical sequences seems to be a problem for updating with files. 
 Ideally, we'd catch the duplicate sequence problem before they get stored in the db.

Let me know if I'm not understanding.

src/range.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    pub fn translate_index(
+        &self,
+        index: i64,
+        range2: &Range,
+        sequence_length: i64,
+        is_circular_contig: bool,
+    ) -> i64 {
+        if !self.contains(index) {
+            panic!("Index {} is not contained in range {:?}", index, self);
```
Chris7, line 187: maybe make this return `Result`? like `Result<i64, &'static str>`. So this condition would be return `Err("Index xxx` and we don't break the api.

src/models/path.rs
```diff
@@ -283,14 +291,231 @@ impl Path {
...
+                            their_block_index += 1;
+                        } else {
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        mappings
```
Chris7, line 420: i'm not sure if we should be returning extra info w/ the range start. As a user receiving back overlapping coordinates, if i put in a heterogenous set of nodes, how would i use the output? For example, an entire blockgroup or 2 different paths.

Chris7 (approved):
it looks good. Just some questions and thoughts on the data return api.

src/range.rs
dkhofer: Sure, done

src/models/path.rs
dkhofer: I'll be honest, I'm not understanding the question.  I'll start a slack thread

PR #76, opened 2024-10-25
Initial edge updates

dkhofer:


src/main.rs
```diff
@@ -20,7 +22,7 @@ use std::str;
 #[derive(Parser)]
 #[command(version, about, long_about = None)]
 struct Cli {
-    /// The path to the database you wish to utilize
```
dkhofer, line 23: I don't know how these triple slashes got in here, only noticed them with this branch

src/operation_management.rs
```diff
@@ -801,6 +802,41 @@ pub fn move_to(conn: &Connection, operation_conn: &Connection, operation: &Opera
     }
 }
 
+pub fn start_operation<'a>(
```
dkhofer, line 805: Just some refactoring

src/models/path.rs
```diff
@@ -146,6 +147,25 @@ impl Path {
         rows.next().unwrap().unwrap()
     }
 
+    pub fn full_query(conn: &Connection, query: &str, params: impl Params) -> Vec<Path> {
```
Chris7, line 150: What's the difference here and other queries? Does this version support carray?

src/models/path.rs
```diff
@@ -524,6 +544,62 @@ impl Path {
...
+                }
+            } else if after_new_node {
+                new_edge_ids.push(edge.id);
+            } else if edge.id == edge_after_new_node.id {
+                after_new_node = true;
+                new_edge_ids.push(edge.id);
+            }
+        }
+
+        let new_name = format!(
```
Chris7, line 597: i was thinking the fasta header should be part of the new name. So it would be like `chr1-GFP`, etc.

src/updates/fasta.rs
```diff
@@ -0,0 +1,509 @@
...
+    node::Node,
+    operation_path::OperationPath,
+    path::{Path, PathBlock},
+    sequence::Sequence,
+    strand::Strand,
+    traits::*,
+};
+use crate::{calculate_hash, operation_management};
+
+pub fn update_with_fasta(
```
Chris7, line 19: Is this compatible as a branch off of #79 ? Where `insert_bg_changes` is the way to give updates? We're quickly moving to the multiple engineering round setups where this is still limited to 1 round. (e.g. we do one round of changes to insert a landing pad, and then another round to insert many genes into it. We can't refer to the landing pad region here).

dkhofer: Let's say some fasta was imported by operation 1, and the region we want to replace with a landing pad is 10-20 on chr1.

```
gen update --fasta landing-pad.fa --start 10 --end 20 --contig-name chr1
```

Let's say that gets an operation ID of 2.  This creates a new path for chr1 and adds the non-chr1 paths from operation 1, plus the new chr1, as entries in operation_paths associated with this operation 2.  That means the current chr1 path (with landing pad) is what the next operation will reference for coordinates.

Let's say we want to integrate a gene to the landing pad starting at 12, going to 100.  (No idea how biologically valid that is.)

```
gen update --fasta landing-pad.fa --start 12 --end 100 --contig-name chr1
```

That creates another new path, replacing the one that was created for operation 2.  All the non-chr1 paths stay the same from operation 1.

Does that help explain my thinking?

src/models/path.rs
dkhofer: It does, but it turns out I don't need that for this PR.  I'll put together a separate PR with this change for the places that need it

src/updates/fasta.rs
```diff
@@ -0,0 +1,587 @@
...
+    for block_group in block_groups {
+        let new_bg_id = BlockGroup::get_or_create_sample_block_group(
+            conn,
+            collection_name,
+            new_sample_name,
+            &block_group.name,
+            Some(parent_sample_name),
+        )
+        .unwrap();
+        BlockGroup::clone(conn, block_group.id, new_bg_id);
```
Chris7, line 62: what's the reason for the second clone?

src/updates/fasta.rs
```diff
@@ -0,0 +1,587 @@
...
+        if block_group.name == region_name {
+            new_block_group_id = new_bg_id;
+        }
+    }
+
+    if new_block_group_id == 0 {
+        panic!("No block group found with region name: {}", region_name);
+    }
+
+    let path = BlockGroup::get_current_path(conn, new_block_group_id);
```
Chris7, line 72: we should ensure the clone in blockgroup orders the newly created paths by order of id too so this actually returns the first path made in a cloned bg.

src/updates/vcf.rs
```diff
@@ -243,7 +244,7 @@ pub fn update_with_vcf<'a>(
                 collection_name,
                 &fixed_sample,
                 seq_name.clone(),
-                coordinate_frame,
+                Some(""),
```
Chris7, line 247: what's the reason for these? it should already be passing None

Chris7 (approved):
Just a few comments worth checking on. No need to send back if things are ok.

src/updates/fasta.rs
dkhofer: Oh, I didn't see that get_or_create_sample_block_group did the clone.  Removed this line, thanks!

src/updates/vcf.rs
dkhofer: Long story short, I was confused about our sample name conventions.  I've reverted most of the changes to this file

src/updates/fasta.rs
dkhofer: Just made that change

PR #79, opened 2024-10-29
WIP: Blockgroup modifications

Chris7:
Enable modifications directly on blockgroups instead of paths


src/models/block_group.rs
```diff
@@ -606,13 +689,268 @@ impl BlockGroup {
 
         new_edges
     }
+
+    pub fn insert_bg_changes(conn: &Connection, changes: &Vec<PathChange>) {
```
dkhofer, line 693: These methods being added are almost identical to existing ones (without the `_bg` in the name).  Could you just modify those methods instead?

src/updates/vcf.rs
```diff
@@ -246,8 +252,13 @@ pub fn update_with_vcf(
                     let allele_accession = accession_name
                         .clone()
                         .filter(|name| gt.allele as i32 == accession_allele);
+                    let mut ref_start = (record.variant_start().unwrap().unwrap().get() - 1) as i64;
                     if gt.allele != 0 {
-                        let alt_seq = alt_alleles[chromosome_index - 1];
+                        let mut alt_seq = alt_alleles[chromosome_index - 1];
+                        if alt_seq != "*" && alt_seq.len() < ref_seq.len() {
```
Chris7, line 258: so this tried to detect a deletion and offsets the position by 1 if it's a deletion. This is so we don't get the reference base made as a deletion node and the deletions look like clean edges instead of connecting through a reference base.

src/imports/fasta.rs
```diff
@@ -133,7 +133,7 @@ mod tests {
         setup_gen_dir();
         let mut fasta_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         fasta_path.push("fixtures/simple.fa");
-        let conn = get_connection(None);
+        let conn = get_connection("taf.db");
```
dkhofer, line 136: What is taf.db?

dkhofer:
Mostly looks good, I'm still trying to understand some of the details, asked some questions in slack

src/models/block_group.rs
Chris7: yeah, i wanted to establish that we agreed on this before.

src/graph.rs
```diff
@@ -87,6 +155,37 @@ mod tests {
...
+            GraphEdge {
+                edge_id: 1,
+                chromosome_index: 1,
+                phased: 1,
+                source_strand: Strand::Forward,
+                target_strand: Strand::Reverse,
+            },
+        );
+        let x = all_simple_paths_by_edge(&graph, a, b).next().unwrap();
+        println!("{x:?}");
```
dkhofer, line 186: Should there be asserts here?

dkhofer:
I think I understand enough that I'm going to approve.

One suggestion: I think `prune_graph` could be moved alongside `build_graph` in the Edge model, and `prune` could be a parameter to that method.  Pretty much everywhere `prune_graph` is called, is right after `build_graph` in Edge or `get_graph` in BlockGroup

dkhofer (approved):

PR #89, opened 2024-11-12
Add command to propagate annotations

dkhofer:


src/annotations/gff.rs
```diff
@@ -0,0 +1,184 @@
...
+        .map(|bg| (bg.name.clone(), BlockGroup::get_current_path(conn, bg.id)))
+        .collect::<HashMap<String, Path>>();
+    let target_paths_by_name = target_block_groups
+        .iter()
+        .map(|bg| (bg.name.clone(), BlockGroup::get_current_path(conn, bg.id)))
+        .collect::<HashMap<String, Path>>();
+
+    let mut path_mappings_by_name = HashMap::new();
+    for (name, target_path) in target_paths_by_name.iter() {
+        let source_path = source_paths_by_name.get(name).unwrap();
```
Chris7, line 40: why do we assume the 2 sample groups need the same paths?

src/annotations/gff.rs
```diff
@@ -0,0 +1,184 @@
...
+            .set_type(record.ty().to_string())
+            .set_start(
+                Position::new(propagated_annotation.start.try_into().unwrap())
+                    .expect("Could not convert start ({start}) to usize for propagation"),
+            )
+            .set_end(
+                Position::new(propagated_annotation.end.try_into().unwrap())
+                    .expect("Could not convert end ({end}) to usize for propagation"),
+            )
+            .set_strand(record.strand())
```
Chris7, line 77: i think we're missing some columns like phase

src/annotations/gff.rs
```diff
@@ -0,0 +1,184 @@
...
+                assert_eq!(record.end().get(), 26);
+            } else {
+                assert_eq!(record.reference_sequence_name(), "m123");
+                assert_eq!(record.source(), "gen-test");
+                assert_eq!(record.ty(), "Gene");
+                // Gene annotation, was on (5, 20)
+                // Replaced (15, 25) with a 2 bp sequence
+                // New gene annotation is (5, 15)
+                assert_eq!(record.start().get(), 5);
+                assert_eq!(record.end().get(), 15);
```
Chris7, line 180: what happens if we throw an insert into a gene? wondering if we split it up.

Chris7 (approved):
Only question is about requiring paths to match. I think that lookup can be on-demand cache in an anonymous function.

src/annotations/gff.rs
dkhofer: We "bridge" the annotation across the split up parts.  The first check is on a region-wide annotation and confirms it still spans the region across a replaced part.  I edited the comments to make that clearer.  If you want a check on a gene annotation specifically, I can put one in

src/annotations/gff.rs
dkhofer: The paths are associated with block group name (eg "chr22").  I just changed the variable names to make that clearer (eg `source_paths_by_name` -> `source_paths_by_bg_name`).  So the two associated paths are the latest paths for block groups with the same name.  Those pairs of paths are what we propagate annotations on.  Let me know if I can do more to clarify that

PR #97, opened 2024-11-21
Patch import export

Chris7:


src/graph.rs
```diff
@@ -31,6 +32,63 @@ pub struct GraphEdge {
     pub target_strand: Strand,
 }
 
+#[derive(Debug)]
+pub struct OperationGraph {
```
Chris7, line 36: We can't use a strings for node ids, so this structure provides a mapping between operation hashes and graph node ids.

src/operation_management.rs
```diff
@@ -741,46 +738,48 @@ pub fn reset(conn: &Connection, operation_conn: &Connection, db_uuid: &str, op_i
...
 
-pub fn merge(
+pub fn merge<'a>(
     conn: &Connection,
     operation_conn: &Connection,
     db_uuid: &str,
     source_branch: i64,
     other_branch: i64,
-) {
+    force_hash: impl Into<Option<&'a str>>,
```
Chris7, line 779: this is so we can force hashes for testing or in the future maybe empty commits

migrations/operations/01-initial/up.sql
```diff
@@ -6,25 +6,22 @@ CREATE TABLE defaults (
...
+  parent_hash TEXT,
   branch_id INTEGER NOT NULL,
-  collection_name TEXT,
   change_type TEXT NOT NULL,
   change_id INTEGER NOT NULL,
-  FOREIGN KEY(parent_id) REFERENCES operation(id)
+  FOREIGN KEY(parent_hash) REFERENCES operation(hash)
   FOREIGN KEY(branch_id) REFERENCES branch(id)
 ) STRICT;
-CREATE UNIQUE INDEX operation_uidx ON operation(db_uuid, branch_id, hash);
```
dkhofer, line 27: Why is this index going away?

src/graph.rs
dkhofer: :+1:

src/operation_management.rs
```diff
@@ -445,9 +451,16 @@ pub fn apply_changeset(conn: &Connection, operation: &Operation) {
...
                     node_map.insert(
                         node_pk,
-                        str::from_utf8(item.new_value(1).unwrap().as_bytes().unwrap())
-                            .unwrap()
-                            .to_string(),
+                        (
+                            str::from_utf8(item.new_value(1).unwrap().as_bytes().unwrap())
+                                .unwrap()
+                                .to_string(),
+                            item.new_value(2)
```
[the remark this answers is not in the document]
dkhofer: I just noticed this today too!  Wasn't sure if it was intentional, glad it's fixed

src/operation_management.rs
dkhofer: :+1:

src/patch.rs
```diff
@@ -0,0 +1,243 @@
...
+            "".to_string(),
+            "".to_string(),
+            conn,
+            operation_conn,
+            None,
+        )
+        .unwrap();
+        let mut write_stream: Vec<u8> = Vec::new();
+        create_patch(operation_conn, &[op_1.hash, op_2.hash], &mut write_stream);
+        load_patches(&write_stream[..]);
```
dkhofer, line 151: Maybe also do a get

dkhofer (approved):
This is very cool, and you were basically just reusing all the branch functionality you already added, you didn't need to add much substantial new logic.  I like the git-style syntax too.  LGTM, comments are non-blocking

migrations/operations/01-initial/up.sql
Chris7: because hash is the primary key, it's unique by default. Took me a bit ot remember why :)

src/patch.rs
```diff
@@ -0,0 +1,243 @@
...
+            conn,
+            operation_conn,
+            None,
+        )
+        .unwrap();
+        let mut write_stream: Vec<u8> = Vec::new();
+        create_patch(operation_conn, &[op_1.hash, op_2.hash], &mut write_stream);
+        let patches = load_patches(&write_stream[..]);
+        apply_patches(conn2, operation_conn2, &patches);
+        apply_patches(conn, operation_conn, &patches);
```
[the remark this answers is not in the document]
Chris7: good idea, i like it

PR #98, opened 2024-11-22
Switch to hash as primary key for operations

Chris7:


src/graph.rs
```diff
@@ -31,6 +32,63 @@ pub struct GraphEdge {
     pub target_strand: Strand,
 }
 
+#[derive(Debug)]
+pub struct OperationGraph {
```
Chris7, line 36: We can't use strings as node ids as they can't be copied, so we make a custom graph strutcure for it to keep the string -> int mapping for nodes.

PR #99, opened 2024-11-22
Add GFA diff operation

dkhofer:


src/diffs/gfa.rs
```diff
@@ -0,0 +1,615 @@
...
+                sequence: source_sequence[start..end].to_string(),
+                node_id: source_block.node_id,
+                sequence_start: source_block.sequence_start,
+                strand: source_block.strand,
+            };
+            segments.insert(segment.clone());
+            source_segment_ids.push(segment.segment_id());
+            source_strands.push(source_block.strand);
+        }
+        for (block1, block2) in source_node_blocks.iter().tuple_windows() {
```
Chris7, line 103: just a heads up that tuple_windows will be effectively a noop on a list of length 1. Will the uses here always have >= 2 items?

src/diffs/gfa.rs
```diff
@@ -0,0 +1,615 @@
...
+use crate::models::{block_group::BlockGroup, path::Path, sample::Sample};
+use crate::range::Range;
+
+pub fn gfa_sample_diff(
+    conn: &Connection,
+    collection_name: &str,
+    filename: &PathBuf,
+    from_sample_name: Option<&str>,
+    to_sample_name: &str,
+) {
```
Chris7, line 18: can you put in a comment the general approach here?

src/diffs/gfa.rs
```diff
@@ -0,0 +1,615 @@
...
+        if mappings.is_empty() {
+            continue;
+        }
+        let mut source_ranges = vec![];
+        let mut target_ranges = vec![];
+
+        let mut source_mapping_start = 0;
+        let mut target_mapping_start = 0;
+        for mapping in &mappings {
+            if mapping.source_range.start > source_mapping_start {
```
Chris7, line 53: can you put in a comment about the source_mapping_start and what is happening here? I'm confusing why we have this.

src/diffs/gfa.rs
Chris7: looking at node_block_parition, looks like this could be a single block.

src/models/path.rs
```diff
@@ -235,29 +242,12 @@ impl Path {
         sequences_by_node_id: &HashMap<i64, Sequence>,
         current_path_length: i64,
     ) -> PathBlock {
-        if into.target_node_id != out_of.source_node_id {
```
dkhofer, line 238: These checks aren't needed here any more because of the validations I added

src/diffs/gfa.rs
```diff
@@ -0,0 +1,644 @@
...
+    let target_paths_by_name = target_block_groups
+        .iter()
+        .map(|bg| (bg.name.clone(), BlockGroup::get_current_path(conn, bg.id)))
+        .collect::<HashMap<String, Path>>();
+
+    let mut segments = HashSet::new();
+    let mut links = HashSet::new();
+    let mut paths = vec![];
+
+    for (target_name, target_path) in target_paths_by_name.iter() {
```
Chris7, line 38: if a path gets deleted, wouldn't it not be present here?

src/diffs/gfa.rs
```diff
@@ -0,0 +1,644 @@
...
+            source_ranges.push(mapping.source_range.clone());
+            source_mapping_start = mapping.source_range.end;
+            if mapping.target_range.start > target_mapping_start {
+                target_ranges.push(Range {
+                    start: target_mapping_start,
+                    end: mapping.target_range.start,
+                });
+            }
+            target_ranges.push(mapping.target_range.clone());
+            target_mapping_start = mapping.target_range.end;
```
Chris7, line 68: can we call this something like last_position or something? would make it more obvious what this is in how its used below.

src/diffs/gfa.rs
dkhofer: Done

src/diffs/gfa.rs
dkhofer: It should be fine.  If there is a single block then there is only one segment, and we want to not generate any links in that case

dkhofer: But I appreciate the warning!

src/diffs/gfa.rs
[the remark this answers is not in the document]
dkhofer: I reworked the logic to handle the case where the source or the target has a block group that the other lacks

src/diffs/gfa.rs
dkhofer: Yep, renamed to last_source_position and last_target_position, makes sense

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1058 @@
...
+                phased: 0,
+            })
+            .collect::<Vec<BlockGroupEdgeData>>();
+        BlockGroupEdge::bulk_create(&conn, &block_group_edges);
+
+        let _path1 = Path::create(&conn, "parent", block_group.id, &edge_ids);
+
+        let temp_dir = tempdir().unwrap();
+        let gfa_path = temp_dir.path().join("self-diff.gfa");
+        gfa_sample_diff(&conn, collection_name, &gfa_path, None, "child");
```
Chris7, line 542: i'm confused why this would work when the `child` sample doesn't exist.

dkhofer: I misnamed this test.  It was actually comparing a sample with stuff against the None sample, which didn't have anything (no block groups).  So it's like doing a git diff when you've just created a file and it prints out the whole file.

That said, I also just added a test comparing a sample to itself...

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1058 @@
...
+
+        import_gfa(&gfa_path, "test collection 2", None, &conn);
```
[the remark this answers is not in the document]
dkhofer: I'm...not sure, at least for GFA format.  In general, we provide a GFA with all segments, links, and paths, not just the segments that are different.  My preference is to be consistent with that style even for a self diff.

I do want to put together a PR that prints out a pure sequence diff on the command line, closer to what git does, and I agree in that case, the output should be empty

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+            }
+            target_ranges.push(mapping.target_range.clone());
+            last_target_position = mapping.target_range.end;
+        }
+
+        if has_source_path {
+            let source_path = source_path_result.unwrap();
+            let source_sequence = source_path.sequence(conn);
+
+            let source_len = source_sequence.len().try_into().unwrap();
```
Chris7, line 120: i think you can do this like `xxx.len() as i64`

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+        writer
+            .write_all(&path_line(&path).into_bytes())
+            .unwrap_or_else(|_| panic!("Error writing path {} to GFA stream", path.name));
+    }
+}
+
+fn segments_from_blocks(node_blocks: &Vec<NodeIntervalBlock>, sequence: &str) -> Vec<Segment> {
+    let mut segments = vec![];
+    for block in node_blocks {
+        let start = usize::try_from(block.start).unwrap();
```
Chris7, line 180: `block.start as usize` might be more concise here too

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+            target_strand: block2.strand,
+        };
+        links.push(link);
+    }
+
+    links
+}
+
+fn path_from_segments(sample_name: Option<&str>, path: &Path, segments: &[Segment]) -> GFAPath {
+    let path_name = if sample_name.is_some() && sample_name.unwrap() != "" {
```
Chris7, line 227: maybe do `sample_name.unwrap_or("") != ""`?

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+                "AACCCCAATTGGGGTT",
+                "AAAAAAAATTGGGGTT"
+            ]
+            .iter()
+            .map(|s| s.to_string())
+            .collect::<HashSet<String>>()
+        );
+
+        let gfa_path = temp_dir.path().join("child-grandchild-diff.gfa");
+        gfa_sample_diff(
```
Chris7, line 454: :clap:

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+            ["AACCCCAATTTTTTTT", "AACCCCAATTGGGGTT"]
+                .iter()
+                .map(|s| s.to_string())
+                .collect::<HashSet<String>>()
+        );
+    }
+
+    #[test]
+    fn test_gfa_diff_against_nothing() {
+        // Confirm diff of a sample to itself is just a single segment
```
Chris7, line 481: `sample to nothing`

src/diffs/gfa.rs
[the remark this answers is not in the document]
Chris7: gotcha. I think we'll probably need to refine how it's called. If I do `diff a.txt a.txt`, I expect nothing. I understand why the GFA format is different. It's focused on the edges/etc. between things.

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+            ["AAAAAAAATTTTTTTT"]
+                .iter()
+                .map(|s| s.to_string())
+                .collect::<HashSet<String>>()
+        );
+    }
+
+    #[test]
+    fn test_gfa_diff_unrelated_paths() {
+        // Confirm diff of a sample to itself is just a single segment
```
Chris7, line 661: update test comment info

Chris7: there's a few tests w/ this same info copy/pasted

Chris7 (approved):
Minor stuff. LGTM! Much easier to process once I considered it as a path diff and the goal being a GFA showing deltas.

src/diffs/gfa.rs
dkhofer: Nice, done

src/diffs/gfa.rs
dkhofer: Yes.  Done

src/diffs/gfa.rs
dkhofer: Yep, done

src/diffs/gfa.rs
dkhofer: Fixed

src/diffs/gfa.rs
```diff
@@ -0,0 +1,1157 @@
...
+                phased: 0,
+            })
+            .collect::<Vec<BlockGroupEdgeData>>();
+        BlockGroupEdge::bulk_create(&conn, &block_group_edges);
+
+        let _path1 = Path::create(&conn, "test path", block_group.id, &edge_ids);
+
+        let temp_dir = tempdir().unwrap();
+        let gfa_path = temp_dir.path().join("diff-against-nothing.gfa");
+        gfa_sample_diff(&conn, collection_name, &gfa_path, None, "test sample");
```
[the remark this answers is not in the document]
dkhofer: I had been thinking in terms of "source sample, target sample" but it kind of doesn't matter, so I just changed it so they're both Option<&str>

src/diffs/gfa.rs
dkhofer: Oops, fixed

PR #101, opened 2024-11-25
Add validations on edges when creating a path

dkhofer:
The rules are:

1. No duplicate edges
2. All edges for a path must be in the path's block group
3. The target of one edge and the source of the next one must be the same node
4. The target coordinate of one edge must be before the source coordinate of the next one
5. An edge can't have the same node + coordinate as both its source and target

### DIFF
--- a/src/models/path.rs
+++ b/src/models/path.rs
@@ -1193,12 +1193,7 @@ mod tests {
         let edge_ids = vec![edge3.id, edge4.id, edge5.id];
         BlockGroupEdge::bulk_create(conn, block_group.id, &edge_ids);
 
-        let path2 = Path::create(
-            conn,
-            "chr2",
-            block_group.id,
-            &[edge3.id, edge4.id, edge5.id],
-        );
+        let path2 = Path::create(conn, "chr2", block_group.id, &edge_ids);
 
         assert_eq!(path2.sequence(conn), "ATCGTTTTTTTT");
 
@@ -1582,12 +1577,7 @@ mod tests {
         let edge_ids = vec![edge1.id, edge2.id, edge3.id];
         BlockGroupEdge::bulk_create(conn, block_group.id, &edge_ids);
 
-        let path1 = Path::create(
-            conn,
-            "chr1",
-            block_group.id,
-            &[edge1.id, edge2.id, edge3.id],
-        );
+        let path1 = Path::create(conn, "chr1", block_group.id, &edge_ids);
 
         let sequence3 = Sequence::new()
             .sequence_type("DNA")
@@ -1711,12 +1701,7 @@ mod tests {
         let edge_ids = vec![edge1.id, edge2.id, edge3.id];
         BlockGroupEdge::bulk_create(conn, block_group.id, &edge_ids);
 
-        let path1 = Path::create(
-            conn,
-            "chr1",
-            block_group.id,
-            &[edge1.id, edge2.id, edge3.id],
-        );
+        let path1 = Path::create(conn, "chr1", block_group.id, &edge_ids);
 
         let sequence3 = Sequence::new()
             .sequence_type("DNA")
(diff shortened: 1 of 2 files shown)


src/operation_management.rs
```diff
@@ -599,24 +620,13 @@ pub fn apply_changeset(conn: &Connection, operation: &Operation) {
...
-
-    for (bg_id, edge_id) in insert_block_group_edges {
-        let bg_id = *dep_bg_map
-            .get(&bg_id)
-            .or(blockgroup_map.get(&bg_id).or(Some(&bg_id)))
-            .unwrap();
-        let edge_id = dep_edge_map
-            .get(&edge_id)
-            .or(edge_id_map.get(&edge_id).or(Some(&edge_id)))
+        let new_bg_id = *dep_bg_map
```
dkhofer, line 623: This fixes a bug pointed out by the validation checking if the path edges are in the block group.  We weren't mapping the path's previous block group ID to the new one generated for the changeset

src/models/path.rs
```diff
@@ -84,7 +85,63 @@ pub struct Annotation {
...
+    pub fn validate_edges(conn: &Connection, edge_ids: &[i64], block_group_id: i64) {
+        let edge_id_set = edge_ids.iter().collect::<HashSet<_>>();
+
+        // No duplicate edges allowed
+        if edge_id_set.len() != edge_ids.len() {
+            panic!("Duplicate edge IDs in path creation");
+        }
+
+        // All path edges must be in the path's block group
+        let edges = BlockGroupEdge::edges_for_block_group(conn, block_group_id);
```
Chris7, line 97: what would be the mechanism for cross-bg linkages? For example we have a blockgroup of chromosome 1 and another for chromosome 2 and there's a translocation between the 2.

src/models/path.rs
```diff
@@ -84,7 +85,63 @@ pub struct Annotation {
...
+            .map(|edge| (edge.id, edge))
+            .collect::<HashMap<_, _>>();
+
+        // Two consecutive edges must share a node
+        // Two consecutive edges must not go into and out of a node at the same coordinate
+        for (edge1_id, edge2_id) in edge_ids.iter().tuple_windows() {
+            let edge1 = edges_by_id.get(edge1_id).unwrap();
+            let edge2 = edges_by_id.get(edge2_id).unwrap();
+            if edge1.target_node_id != edge2.source_node_id {
+                panic!(
```
Chris7, line 114: should we handle these with asserts instead of the if/panics? Might condense it and make it more obvious these are validations

src/models/path.rs
```diff
@@ -84,7 +85,62 @@ pub struct Annotation {
...
+                    "Edges {} and {} don't share the same node ({} vs. {})",
+                    edge1.id, edge2.id, edge1.target_node_id, edge2.source_node_id
+                );
+            }
+
+            if edge1.target_coordinate >= edge2.source_coordinate {
```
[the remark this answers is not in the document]
Chris7: i think this should be enforced -- not doing it would be the style where there is no strand and we check for end < start for reverse strand. It becomes a mess.

src/operation_management.rs
Chris7: oof

Chris7 (approved):
Just 2 questions -- not blocking. But if we're not sure, maybe we can put warnings on those 2 cases so we can catch errors easily if its not desired.

src/models/path.rs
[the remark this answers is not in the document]
dkhofer: :+1:

src/models/path.rs
dkhofer: So that's outside the scope of this PR :) but at a minimum,
* they'll share the common sequence
* edges would need to be created into and out of that sequence for 2
* we should copy information from any relevant block group edges from 1 to 2 (chromosome index, phased maybe)
* I don't think paths could be copied

I haven't thought deeply about this, I'm going to defer until we actually implement something that can result in a translocation

src/models/path.rs
```diff
@@ -84,7 +85,63 @@ pub struct Annotation {
 }
 
 impl Path {
+    pub fn validate_edges(conn: &Connection, edge_ids: &[i64], block_group_id: i64) {
+        let edge_id_set = edge_ids.iter().collect::<HashSet<_>>();
+
+        // No duplicate edges allowed
```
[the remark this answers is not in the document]
dkhofer: I changed it to just print the message for duplicate edges, I'm fine with waiting to see if it needs to be stricter in some way

src/models/path.rs
dkhofer: Good point, done

PR #102, opened 2024-11-27
Genbank parser

Chris7:


Cargo.toml
```diff
@@ -18,16 +18,17 @@ itertools = "0.13.0"
...
+rusqlite_migration = { version = "1.3.1" , features = ["from-directory"]}
 serde = {  version = "1.0.210", features = ["derive"] }
 serde_json = "1.0.128"
 sha2 = "0.10.8"
-tempfile = "3.12.0"
+tempfile = "3.14.0"
 interavl = "0.2.0"
 regex = "1.11.1"
 flate2 = "1.0.35"
+gb-io = "0.7.1"
```
dkhofer, line 31: I can't believe noodles doesn't support genbank

src/imports/genbank.rs
```diff
@@ -0,0 +1,356 @@
...
+use regex::Regex;
+use rusqlite::Connection;
+use std::io::Read;
+use std::str;
+
+pub fn import_genbank<'a, R>(conn: &Connection, data: R, collection: impl Into<Option<&'a str>>)
+where
+    R: Read,
+{
+    let reader = reader::SeqReader::new(data);
```
dkhofer, line 20: I know you said the next PR will have operation tracking, but it seems like it would be fairly easy to call start_operation and end_operation in this method?

dkhofer (approved):
LGTM, nothing blocking.  This is all fairly straightforward!

Cargo.toml
Chris7: i was wondering about this as well, and there's actually a reason. noodles is specifically stuff in the hts spec: https://samtools.github.io/hts-specs/

PR #103, opened 2024-11-27
Chromosome fields to block group edge

dkhofer:


src/models/block_group_edge.rs
```diff
@@ -1,14 +1,39 @@
...
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct BlockGroupEdgeData {
+    pub block_group_id: i64,
+    pub edge_id: i64,
+    pub chromosome_index: i64,
+    pub phased: i64,
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct AugmentedEdge {
```
dkhofer, line 26: I needed to add this (and the Data version) in order to join an edge with the chromosome index/phased info in memory

src/operation_management.rs
```diff
@@ -1491,15 +1517,24 @@ mod tests {
             rusqlite::params!(),
         )
         .len();
-        // NOTE: The edge count is 10 because of the following:
+        // NOTE: 3 block groups get created with the update from vcf, corresponding to the unknown, G1, and foo samples
```
dkhofer, line 1520: This test broke for various reasons, so I went through and looked more carefully at the data being created.  Let me know if any of this is unexpected

Chris7 (approved):
lgtm

PR #104, opened 2024-12-03
Add operation for genbank

Chris7:
This is why I wanted a separate PR. The signature for import_genbank is a reader -- not a file path so operations couldn't be made. So I added a method to add operation info if it should be tracked. The reason for this is I felt like our import operations are a bit too constrained to create apis around. Accepting a stream of input makes it easier for supporting pipes/etc. and other non-file based inputs. Having the caller responsible for telling the function the file name instead of coupling that to the reader method (being a file-like object) gives a bit more flexibility.

I also wanted better errors here, but it required refactoring a bunch of other error calls.

### DIFF
--- a/src/imports/genbank.rs
+++ b/src/imports/genbank.rs
@@ -230,7 +230,7 @@ mod tests {
     use crate::models::file_types::FileTypes;
     use crate::models::metadata;
     use crate::models::operations::setup_db;
-    use crate::test_helpers::{get_connection, get_operation_connection};
+    use crate::test_helpers::{get_connection, get_operation_connection, setup_gen_dir};
     use std::collections::HashSet;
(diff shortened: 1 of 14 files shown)


src/imports/genbank.rs
```diff
@@ -186,14 +210,24 @@ where
...
-            Err(e) => println!("Failed to parse {e:?}"),
+            Err(e) => return Err(GenBankError::ParseError(format!("Failed to parse {}", e))),
         }
     }
+    end_operation(
+        conn,
+        op_conn,
+        &mut session,
+        operation_info,
+        "something",
```
dkhofer, line 221: This probably needs to change

dkhofer (approved):
LGTM!  One minor comment

PR #105, opened 2024-12-04
Add a couple commands to list and show basic stuff

dkhofer:


Cargo.toml
```diff
@@ -21,6 +21,7 @@ petgraph = "0.6.5"
 remove_dir_all = "1.0.0"
 rusqlite = { version = "0.32.1", features = ["bundled", "array", "session"] }
 rusqlite_migration = { version = "1.3.1" , features = ["from-directory"]}
+ruzstd = "0.7.3"
```
dkhofer, line 24: This is to resolve https://github.com/ginkgobioworks/gen/security/dependabot/4

It seemed serious enough to add it to the toml

src/main.rs
```diff
@@ -664,5 +692,44 @@ fn main() {
...
+            let name = &name.clone().unwrap_or_else(|| {
+                get_default_collection(&operation_conn)
+                    .expect("No collection specified and default not setup.")
+            });
+            let block_groups = Sample::get_block_groups(&conn, name, Some(sample));
+            for block_group in block_groups {
+                println!("{}", block_group.name);
+            }
+        }
+        Some(Commands::ShowSequence {
```
Chris7, line 711: i would think `GetX` instead of `ShowX` for general usage, but just an opinion. Like `GetGraph`, etc.

Chris7 (approved):
lgtm. just a comment on show vs get

src/main.rs
Chris7: maybe do region parsing too?

dkhofer: Sure, made both changes

PR #106, opened 2024-12-05
Genbank export

Chris7:
Additions:
* Genbank exports

Changes:
* I changed the way we do imports in this as well. A genbank file will have all its changes to the base sequenced baked into the sequence present, so we now undo all these changes to get back to the initial sequence. This will also make it possible to update from a genbank file as we can now match sequences.
* I briefly was using the graph interval tree as an approach so I moved it to the graph module so any graph can be flattened into an interval tree. I ended up not using this approach but it felt better to have it possible for other graphs to be flattened.


src/models/block_group.rs
```diff
@@ -744,113 +743,7 @@ impl BlockGroup {
         // make a tree where every node has a span in the graph.
         let mut graph = BlockGroup::get_graph(conn, block_group_id);
         BlockGroup::prune_graph(&mut graph);
-        #[derive(Clone, Debug, Ord, PartialOrd, Eq, Hash, PartialEq)]
```
Chris7, line 747: this is a copy/paste to move this method so it can be called by other graphs. it was from an approach i discarded but thought it was better as a function so kept the change.

src/graph.rs
```diff
@@ -194,6 +196,119 @@ where
     reachable
 }
 
+pub fn flatten_to_interval_tree(
```
Chris7, line 199: this is a copy/paste to move a method so it can be called by other graphs. it was from an approach i discarded but thought it was better as a function so kept the change.

src/exports/genbank.rs
```diff
@@ -0,0 +1,833 @@
...
+    // may not be the ones we want, however, as nodes can be reused in a graph. So we traverse
+    // the graph, and match nodes along with the reduced graph. If the traversed set of nodes
+    // is an exact match for the path, we return it. Note there can possibly be multiple traversals
+    // that satisfy the path, but we just return 1.
+
+    // first, reduce the graph down to nodes and connections that are possible
+    let mut path_nodes = HashSet::new();
+    let mut path_edges = HashSet::new();
+    let mut path_it = path_blocks.iter().peekable();
+    while let Some(block) = path_it.next() {
```
dkhofer, line 55: Given that the blocks are coming from the Path `blocks` method, I think this code could be simplified some.  The blocks are going to be in order, and the only terminal ones will be the first and last.  So you could just iterate over a list excluding the first and last items to get the nodes, and then use tuple_windows to get edges.

My suggestion would be to add an `include_terminal` parameter to the Path blocks method that only adds the terminal blocks if true.  You would then have it be false when calling it in this file, but true everywhere else in the codebase, then you don't need to worry about handling the terminal cases here.

src/exports/genbank.rs
```diff
@@ -0,0 +1,833 @@
...
+            }
+        }
+    }
+
+    let blocks = path_blocks
+        .iter()
+        .filter(|block| !Node::is_terminal(block.node_id))
+        .sorted_by(|a, b| Ord::cmp(&a.path_start, &b.path_start))
+        .collect::<Vec<&PathBlock>>();
+    let (_, first_block) = path_blocks
```
dkhofer, line 93: Also not sure why this isn't just `path_blocks.first().unwrap()`

src/imports/genbank.rs
```diff
@@ -107,117 +207,83 @@ where
...
+                            start,
+                            end,
+                            block: PathBlock {
+                                id: 0,
+                                node_id: change_node,
+                                block_sequence: (**new_seq).clone(),
+                                sequence_start: 0,
+                                sequence_end: change_seq.length,
+                                path_start: start,
+                                // TODO -- do i do end + len or is this the position on path we are changing?
```
dkhofer, line 256: Pretty sure it's correct as you have it, but I also think insert_changes doesn't use it.  The field is used by other stuff though

dkhofer (approved):
Overall looks good, going to approve.  I think the code using the path blocks can be simplified, let me know if my suggestions don't work

src/models/block_group.rs
dkhofer: :+1:

src/exports/genbank.rs
Chris7: tuple_windows won't iterate if it's length 1 :(

src/exports/genbank.rs
Chris7: yeah this can all be simpler. thanks

PR #107, opened 2024-12-09
Update from genbank files

Chris7:


examples/externally_edited_files/geneious_engineering.md
```diff
@@ -0,0 +1,35 @@
+# Simple editing workflow with Geneious
+
+This shows a simple back-and-forth editing example using Geneious. Here, sequences are managed in Geneious and exported
+as GenBank files to faciliate analyses.
+
+```console
```
dkhofer, line 6: Could you give a sentence or two of context for what is being downloaded?

src/imports/genbank.rs
```diff
@@ -166,46 +169,61 @@ where
...
-                BlockGroupEdge::bulk_create(conn, &new_block_group_edges);
-                let path = Path::create(
-                    conn,
-                    contig,
-                    block_group.id,
-                    &[edge_into.id, edge_out_of.id],
-                );
+
+                let (path, block_group) = if update && !create_missing {
+                    let block_group = BlockGroup::get(conn, "select * from block_groups where collection_name = ?1 AND sample_name is null AND name = ?2", params![Value::from(collection.name.clone()), Value::from(contig.clone())]).unwrap_or_else(|_| panic!("No block group named {contig} exists. Try importing first or pass --create-missing."));
```
dkhofer, line 174: I think since we have a fair amount of structure to the project now, it would make sense to try to follow a policy of no raw sql outside of model classes.  There may need to be exceptions, but this is a pretty straightforward lookup, and I think it'd be good to add a lookup method to BlockGroup and call it here

src/imports/genbank.rs
```diff
@@ -166,46 +169,61 @@ where
...
+                    let paths = Path::query(
```
dkhofer, line 175: Similarly, it'd be great if you could add a path lookup method and call it here

Chris7:
i can put a refactor on the todo. I put it in import because the 2 approaches are essentially the same. The difference is how paths/blockgroups are gathered.

examples/externally_edited_files/geneious_engineering.md
Chris7: yeah, this is a copy of bob's other example.

src/imports/genbank.rs
Chris7: i'm going to fiddle with something to make the way we query models have less sql and try to update this to it.

examples/externally_edited_files/geneious_engineering.md
dkhofer: :+1:

src/updates/genbank.rs
```diff
@@ -0,0 +1,437 @@
...
+                .map(|s| s.to_lowercase())
+                .collect();
+            let mod_seq = str::from_utf8(&f[1].seq).unwrap().to_string();
+            assert!(sequences.contains(&unchanged_seq));
+            assert!(sequences.contains(&mod_seq));
+        }
+
+        #[test]
+        fn test_errors_on_missing_locus() {
+            // This tests that we are able to take a genbank that has been further modified
```
dkhofer, line 396: I don't think this comment is accurate?  Seems like copypasta

src/updates/genbank.rs
```diff
@@ -0,0 +1,437 @@
...
+            let sequences: HashSet<String> = BlockGroup::get_all_sequences(conn, 1, false)
+                .iter()
+                .map(|s| s.to_lowercase())
+                .collect();
+            let unchanged_seq = get_unmodified_sequence();
+            assert!(sequences.contains(&unchanged_seq));
+            let mod_seq = str::from_utf8(&f[0].seq).unwrap().to_string();
+            assert!(sequences.contains(&mod_seq));
+
+            // we have a new blockgroup called deletion thta uses the same base sequence but
```
dkhofer, line 383: Typo, thta -> that

dkhofer (approved):
Minor comments, otherwise LGTM.  Thanks!

PR #108, opened 2024-12-09
Fix some issues with update with library

dkhofer:
Fixes a bug that was preventing a part from being used in multiple buckets in "update with library".  I introduced the exact same bug in the import library code, so I'm fixing that here too.


src/exports/gfa.rs
```diff
@@ -55,9 +55,8 @@ pub fn export_gfa(
     let boundary_edges = Edge::boundary_edges_from_sequences(&blocks);
     edges.extend(boundary_edges.clone());
 
-    let (mut graph, _edges_by_node_pair) = Edge::build_graph(&edges, &blocks);
-
-    BlockGroup::prune_graph(&mut graph);
+    let (graph, _edges_by_node_pair) = Edge::build_graph(&edges, &blocks);
+    //    BlockGroup::prune_graph(&mut graph);
```
dkhofer, line 59: This was causing problems because all the edges are currently getting a chromosome index of 0 when we update with a library, so most edges were getting pruned.  I'm going to work on the real solution next

src/updates/library.rs
```diff
@@ -94,12 +83,29 @@ pub fn update_with_library(
...
             if !part.is_empty() {
-                let part_id = node_ids_by_name.get(part).unwrap();
-                parts_by_index.entry(index).or_insert(vec![]).push(part_id);
+                let part_hash = sequence_hashes_by_name.get(part).unwrap();
+                let seq_length = sequence_lengths_by_hash.get(part_hash).unwrap();
+                let part_node_id = Node::create(
+                    conn,
+                    part_hash,
+                    calculate_hash(&format!(
+                        "{path_id}:{ref_start}-{ref_end}->{sequence_hash}-column-{index}",
```
dkhofer, line 97: Note that I needed to append `column-{index}` to the end of the hashed string in order to ensure a new node is created for a redundant part that's in a different column

src/imports/library.rs
```diff
@@ -57,19 +57,9 @@ pub fn import_library<'a>(
...
-                "{region_name}:{name}:{ref_start}-{ref_end}->{sequence_hash}",
-                ref_start = 0,
-                ref_end = seq.length,
-                sequence_hash = seq.hash
-            )),
-        );
 
-        node_ids_by_name.insert(name, node_id);
-        sequence_lengths_by_node_id.insert(node_id, seq.length);
+        sequence_hashes_by_name.insert(name, seq.hash.clone());
```
Chris7, line 61: should we check for any overwrites here?

Chris7: or maybe have a better scoped part name? Like for example if someone put `spacer` in multiple buckets and te sequence was different.

src/updates/library.rs
```diff
@@ -113,20 +119,45 @@ pub fn update_with_library(
...
+    };
     let node_start_coordinate = start_coordinate - start_block.start + start_block.sequence_start;
 
-    let end_blocks: Vec<_> = path_intervaltree
-        .query_point(end_coordinate)
-        .map(|x| &x.value)
-        .collect();
-    assert_eq!(end_blocks.len(), 1);
-    let end_block = end_blocks[0];
+    let path_length = path.sequence(conn).len().try_into().unwrap();
```
Chris7, line 142: i have a new `length` method here that might be faster to get this.

Chris7 (approved):
LGTM!

src/imports/library.rs
dkhofer: I checked for overwrites, now it panic!s if there's a duplicate name somewhere in the fasta.  Not going to try to be smarter than that, it's something I think the user should fix if it happens

src/imports/library.rs
```diff
@@ -274,4 +281,87 @@ mod tests {
...
+    #[test]
+    fn one_column_of_parts() {
+        setup_gen_dir();
+        let conn = &get_connection(None);
+        let db_uuid = metadata::get_db_uuid(conn);
+        let op_conn = &get_operation_connection(None);
+        setup_db(op_conn, &db_uuid);
+        let collection = "test";
+
+        let mut parts_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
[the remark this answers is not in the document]
dkhofer: Oh geez, I was using the 2 line version everywhere.  Changed it in these files anyway

src/updates/library.rs
dkhofer: I'm not seeing a `length` method, but I did change this to `path.sequence(conn).len() as i64`, which is a bit simpler

PR #110, opened 2024-12-13
Set up and use phase layers to represent unphased alleles and pooled …

dkhofer:
…design parts


src/config.rs
```diff
@@ -53,7 +53,7 @@ pub fn get_gen_dir() -> String {
         };
         gen_path = cur_dir.join(".gen");
     }
-    return gen_path.to_str().unwrap().to_string();
```
dkhofer, line 56: Clippy got after me for a few things unrelated to this change, but they seemed reasonable

PR #111, opened 2024-12-16
Improve checkout command

Chris7:
Makes the checkout command more git like.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -643,8 +643,8 @@ fn main() {
         Some(Commands::Checkout { branch, hash }) => {
             if let Some(name) = branch.clone() {
                 if Branch::get_by_name(&operation_conn, &db_uuid, &name).is_none() {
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -641,7 +641,37 @@ fn main() {
             operation_management::apply(&conn, &operation_conn, hash, None);
         }
         Some(Commands::Checkout { branch, hash }) => {
-            operation_management::checkout(&conn, &operation_conn, &db_uuid, branch, hash.clone());
+            if let Some(name) = branch.clone() {
+                if Branch::get_by_name(&operation_conn, &db_uuid, &name).is_none() {
+                    println!("Creating branch {name}");
```
dkhofer, line 646: Nitpick but maybe print this after creation, and change "Creating" to "Created"?

dkhofer (approved):
LGTM

PR #112, opened 2024-12-16
Add default gen db and collection

Chris7:


src/updates/genbank.rs
```diff
@@ -277,7 +277,6 @@ mod tests {
     mod geneious_genbanks {
         use super::*;
         use crate::imports::genbank::import_genbank;
-        use rusqlite::ToSql;
```
Chris7, line 280: warning supression

dkhofer: Thanks!

dkhofer (approved):
The water washes over the pebble, gradually making it smooth

PR #114, opened 2024-12-17
Some bug fixes discovered during demo prep

dkhofer:


src/main.rs
```diff
@@ -263,15 +263,15 @@ enum Commands {
         name: Option<String>,
         /// The name of the sample to list graphs for
         #[arg(short, long)]
-        sample: String,
+        sample: Option<String>,
```
dkhofer, line 266: I'm going with having lack of sample name mean the default one is intended

src/models/block_group.rs
```diff
@@ -291,7 +291,11 @@ impl BlockGroup {
             }
         }
         if bg_id == 0 {
-            return Err("No base path exists");
+            let error_string = format!(
+                "No base path exists for sample {} and graph name {}",
+                sample_name, group_name
+            );
+            return Err(Box::leak(error_string.into_boxed_str()));
```
dkhofer, line 298: This was a hoop I had not had to jump through before, it's because the result error return type is static str

src/operation_management.rs
```diff
@@ -757,17 +757,15 @@ pub fn reset(conn: &Connection, operation_conn: &Connection, db_uuid: &str, op_h
...
     }
-    move_to(
-        conn,
-        operation_conn,
-        &Operation::get_by_hash(operation_conn, op_hash)
-            .unwrap_or_else(|_| panic!("Hash {op_hash} does not exist.")),
-    );
+    let operation = Operation::get_by_hash(operation_conn, op_hash)
+        .unwrap_or_else(|_| panic!("Hash {op_hash} does not exist."));
+    let full_op_hash = operation.hash.clone();
```
dkhofer, line 762: This makes things work right if the input hash is a prefix of the full hash

src/main.rs
Chris7: yeah i like the no sample -> whatever was the first import convention.

src/models/block_group.rs
Chris7: for these, we need to start defining our own error codes (see the stuff like GenBankError), then we can use map_err to transform these.

Chris7 (approved):
LGTM!

PR #115, opened 2024-12-21
Add command to update with GFA

dkhofer:


src/gfa_reader.rs
```diff
@@ -158,8 +158,8 @@ pub struct Link<T: SampleType, S: Opt, U: Opt> {
 #[derive(Debug, Clone, Ord, Eq, PartialOrd, PartialEq)]
 pub struct Path<T: SampleType, S: Opt, U: Opt> {
     pub name: String,
-    pub dir: Vec<bool>,
-    pub nodes: Vec<T>,
+    pub strands: Vec<bool>,
```
dkhofer, line 161: Felt like renaming some of these fields to be clearer

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+        .collect();
+
+    // Find which incoming paths match existing paths, and store information about them and their
+    // segments
+    let mut existing_path_ids_by_new_path_name = HashMap::new();
+    let mut path_segments_by_name = HashMap::new();
+    let mut path_strands_by_name = HashMap::new();
+    for path in &gfa.paths {
+        let path_name = path.name.clone();
+        path_segments_by_name.insert(path_name.clone(), path.segments.clone());
```
Chris7, line 66: do we need to clone these or can we pass a reference? seems like it can be a problem w/ bigger GFAs

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+        .collect::<HashSet<&str>>();
+
+    // Record unmatched paths and walks, update existing matched ones
+    let mut unmatched_path_names = vec![];
+    let mut matched_path_name_by_segment_id = HashMap::new();
+    for path in &gfa.paths {
+        let path_name = &path.name;
+        if matched_path_names.contains(path_name.as_str()) {
+            for segment_id in path.segments.iter() {
+                matched_path_name_by_segment_id.insert(segment_id, path_name);
```
Chris7, line 132: couldn't a segment be in multiple paths? Should this be a set of paths the segment belongs to

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+        .iter()
+        .cloned()
+        .collect::<HashSet<&str>>();
+
+    // Record unmatched paths and walks, update existing matched ones
+    let mut unmatched_path_names = vec![];
+    let mut matched_path_name_by_segment_id = HashMap::new();
+    for path in &gfa.paths {
+        let path_name = &path.name;
+        if matched_path_names.contains(path_name.as_str()) {
```
Chris7, line 130: &path_name should work too vs as_str

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    // Build up a new path by merging the shared nodes from the existing path with newly
+    // created nodes
+    let mut existing_path_position = 0;
+    let mut previous_node_id = PATH_START_NODE_ID;
+    let mut previous_node_coordinate = -1;
+    let mut previous_node_strand = Strand::Forward;
+    let mut new_path_edges = vec![];
+    for (i, segment_id) in unmatched_path_segment_ids.iter().enumerate() {
+        if existing_path_ranges_by_segment_id.contains_key(segment_id) {
```
Chris7, line 260: i think you can do something like `if let Some(start, end) = existing_path_ranges_by_segment_id.get(segment_id)` to avoid the double lookup.

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+        } else {
+            // Current segment is new.  Create a sequence and node for it, then add an edge to the
+            // new node
+            let segment = segments_by_id.get(segment_id).unwrap();
+            let segment_sequence = segment.sequence.get_string(&gfa.sequence);
+            let sequence = Sequence::new()
+                .sequence_type("DNA")
+                .sequence(segment_sequence)
+                .save(conn);
+            let node_id = Node::create(conn, &sequence.hash, None);
```
Chris7, line 300: we should make a node hash so this can be repeated without infinite updates.

src/updates/gfa.rs
```diff
@@ -0,0 +1,523 @@
...
+            ),
+        );
+        assert_eq!(block_groups.len(), 1);
+        assert_eq!(
+            BlockGroup::get_all_sequences(conn, block_groups[0].id, false),
+            HashSet::from_iter(expected_sequences),
+        );
+
+        let temp_dir = tempdir().unwrap();
+        let gfa_diff_path = temp_dir.path().join("parent-child-diff.gfa");
```
Chris7, line 438: could we save this as a general fixture to have a test where we just have the initial simple.fa and no other intermediate assets the alternative paths could maybe be coming from?

Chris7: cool, i gues that is the walk-diff.

src/updates/gfa.rs
dkhofer: Good point, changed to reference

src/updates/gfa.rs
dkhofer: Oops, yep, fixed

src/updates/gfa.rs
dkhofer: Yep, done

src/updates/gfa.rs
dkhofer: Done, but let me know if the hash string could be better

src/updates/gfa.rs
dkhofer: Sure, did this for the path diff too

src/updates/gfa.rs
```diff
@@ -0,0 +1,495 @@
...
+            let segment = segments_by_id.get(segment_id).unwrap();
+            let segment_sequence = segment.sequence.get_string(&gfa.sequence);
+            let sequence = Sequence::new()
+                .sequence_type("DNA")
+                .sequence(segment_sequence)
+                .save(conn);
+            let node_id = Node::create(
+                conn,
+                &sequence.hash,
+                calculate_hash(&format!("{segment_id}_{hash}", hash = &sequence.hash)),
```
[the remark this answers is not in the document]
dkhofer: There's a chicken-and-egg problem because we don't have the path ID yet, we need the nodes + edges between them in order to create the path.  I added the path name here instead, that OK?

src/updates/gfa.rs
```diff
@@ -140,7 +143,10 @@ pub fn update_with_gfa(
         let walk_name = &walk.sample_id;
         if matched_path_names.contains(walk_name.as_str()) {
             for segment_id in walk.segments.iter() {
-                matched_path_name_by_segment_id.insert(segment_id, walk_name);
+                matched_path_name_by_segment_id
+                    .entry(segment_id)
+                    .or_insert(HashSet::new())
```
Chris7, line 148: can we do `or_insert_with(HashSet::new)` to avoid a hash creation on every cycle?

Chris7 (approved):
LGTM! One minor comment w/ on_insert. It'll always execute every loop

src/updates/gfa.rs
dkhofer: Yes.  Good to know, thanks!

PR #118, opened 2025-01-02
Setup progress bar

Chris7:
Added progress bars for all the imports and vcf updates. The other methods i wasn't sure if it made sense since they are kind of instant so a print message might be all we need.

image
image
image
image


dkhofer (approved):
Nice! LGTM

PR #120, opened 2025-01-08
View patches

Chris7:
This will generate a .dot file from a patch. It's some of the starting work to explore how we can show changes to users in reviews, etc

### DIFF
--- a/src/operation_management.rs
+++ b/src/operation_management.rs
@@ -151,11 +151,10 @@ pub fn get_changeset_dependencies(conn: &Connection, mut changes: &[u8]) -> Vec<
                     let edge_pk = item.new_value(pk_column).unwrap().as_i64().unwrap();
                     let source_node_id = item.new_value(1).unwrap().as_i64().unwrap();
                     let target_node_id = item.new_value(4).unwrap().as_i64().unwrap();
-                    println!("new edge {source_node_id} {target_node_id}");
                     created_edges.insert(edge_pk);
                     let nodes = Node::get_nodes(conn, &[source_node_id, target_node_id]);
                     for node in nodes.iter() {
-                        if !created_nodes.contains(&node.id) {
+                        if !created_nodes.contains(&node.id) && !Node::is_terminal(node.id) {
                             previous_sequences.insert(node.sequence_hash.clone());
                             previous_nodes.insert(node.id);
                         }
@@ -430,7 +429,7 @@ pub fn apply_changeset(
 
     let mut dep_node_map = HashMap::new();
     for node in dependencies.nodes.iter() {
-        let new_node_id = Node::create(conn, &node.sequence_hash.clone(), None);
+        let new_node_id = Node::create(conn, &node.sequence_hash, node.hash.clone());
         dep_node_map.insert(&node.id, new_node_id);
     }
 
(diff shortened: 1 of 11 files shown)


src/operation_management.rs
```diff
@@ -51,13 +51,22 @@ pub enum FileMode {
...
+    pub block_group: Vec<BlockGroup>,
+    pub nodes: Vec<Node>,
+    pub edges: Vec<Edge>,
+    pub paths: Vec<Path>,
+    pub accessions: Vec<Accession>,
+    pub accession_edges: Vec<AccessionEdge>,
+}
+
+#[derive(Debug)]
+pub struct ChangesetModels {
```
Chris7, line 64: all models aren't in here, should extend when needed.

src/operation_management.rs
```diff
@@ -87,6 +96,7 @@ pub fn get_changeset_dependencies(conn: &Connection, mut changes: &[u8]) -> Vec<
     let mut previous_edges = HashSet::new();
     let mut previous_paths = HashSet::new();
     let mut previous_accessions = HashSet::new();
+    let mut previous_nodes = HashSet::new();
```
Chris7, line 99: this was missing when nodes were added

src/operation_management.rs
```diff
@@ -295,6 +310,106 @@ pub fn load_changeset(operation: &Operation) -> Vec<u8> {
     contents
 }
 
+fn parse_string(item: &ChangesetItem, col: usize) -> String {
```
Chris7, line 313: makes the parsing out of this stuff a lot easier

src/operation_management.rs
dkhofer: :+1:

src/operation_management.rs
dkhofer: :+1:

src/operation_management.rs
```diff
@@ -295,6 +309,106 @@ pub fn load_changeset(operation: &Operation) -> Vec<u8> {
...
+                    });
+                }
+                "nodes" => {
+                    created_nodes.push(Node {
+                        id: parse_number(item, pk_column),
+                        sequence_hash: parse_string(item, 1),
+                        hash: parse_maybe_string(item, 2),
+                    });
+                }
+                "edges" => created_edges.push(Edge {
```
dkhofer, line 381: I didn't realize brackets weren't needed for matching.  Maybe do the same for the other one liners here?

src/views/patch.rs
```diff
@@ -0,0 +1,228 @@
...
+use html_escape;
+use itertools::Itertools;
+use petgraph::graphmap::DiGraphMap;
+use petgraph::Direction;
+use rusqlite::session::ChangesetIter;
+use std::collections::{HashMap, HashSet};
+use std::io::Read;
+
+pub fn view_patches(patches: &[OperationPatch]) -> HashMap<String, HashMap<i64, String>> {
+    // For each blockgroup in a patch, a
```
dkhofer, line 18: Incomplete comment?

dkhofer (approved):
LGTM, comments are minor

src/operation_management.rs
Chris7: sure

PR #123, opened 2025-01-10
Handle CNV notation

Chris7:
This handles the CNV notation in vcf files

### DIFF
--- a/src/updates/vcf.rs
+++ b/src/updates/vcf.rs
@@ -21,6 +21,8 @@ use noodles::vcf::variant::record::samples::series::Value;
 use noodles::vcf::variant::record::samples::Sample as NoodlesSample;
 use noodles::vcf::variant::record::AlternateBases;
 use noodles::vcf::variant::Record;
+use regex;
+use regex::Regex;
 use rusqlite;
 use rusqlite::{types::Value as SQLValue, Connection};
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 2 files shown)


src/updates/vcf.rs
```diff
@@ -726,6 +751,48 @@ mod tests {
...
+            op_conn,
+            None,
+        )
+        .unwrap();
+
+        assert_eq!(
+            BlockGroup::get_all_sequences(conn, get_sample_bg(conn, &collection, "foo").id, true),
+            HashSet::from_iter(vec![
+                "ATCGATCGATCGGATCGGGAACACACAGAGA".to_string(),
+                "ATCGATCGATCGATCATCATCGATCGGGAACACACAGAGA".to_string()
```
dkhofer, line 791: I'm not familiar with CNV.  For my understanding, why do the <CN0>, <CN2>, <CN3>, <CN4> entries result in just one new sequence?

dkhofer (approved):
I'm confused about the unit test result, but otherwise LGTM

src/updates/vcf.rs
Chris7: the alt allele string is <CN0>,<CN2>,<CN3>,<CN4>. I just have 2 of those alleles selected, 1 and 3. This means we have <CN0> and <CN3>. So we should just have 2 sequences. One will be CN0, meaning a deletion, the other CN3 meaning 3 copies of the ref sequence

PR #124, opened 2025-01-10
Minor fixes to docs + drawing logic

bobvh:
3 minor fixes:

- Documentation updates per chat.

- I also saw segment_id being used in your dot generator, I think that was a leftover from the Python prototype. I had used a segment from the GFA vocab, but it's really a block, hence the edit.

- Deletions were drawn as dashed lines that indicate boundary edges, now they're drawn properly. It's still not complete though, because for heterozygous genomes you want to show the boundary edge as being "real" still.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -162,7 +162,8 @@ enum Commands {
         /// The patch name
         #[arg(short, long)]
         name: String,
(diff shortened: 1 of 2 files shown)


Chris7:
thanks. merged it in.

PR #126, opened 2025-01-13
Add command to call gfa diff

dkhofer:


Chris7 (approved):
LGTM

src/main.rs
```diff
@@ -874,5 +890,22 @@ fn main() {
...
+            sample2,
+            gfa,
+        }) => {
+            let name = &name
+                .clone()
+                .unwrap_or_else(|| get_default_collection(&operation_conn));
+            gfa_sample_diff(
+                &conn,
+                name,
+                &PathBuf::from(gfa),
```
Chris7, line 905: no need to do it here, but i've been using the pattern of passing in something that supports Read/Write to my newer methods, making it easier for testing/etc. So a test can pass in a buffer string to write to, and the cli can pass in a file handle.

dkhofer: Is the Read/Write usage in the GAF update code an example?  I think I'm going to skip that here for now.  The pattern of the GFA diff tests is to manually build up stuff in the database, then write to the (temp) diff file, then re-import it and confirm get_all_sequences works as expected.  I could see a buffer string replacing the use of the temp file, but that's about it, so it doesn't seem super compelling.  Is there something I'm missing?

PR #131, opened 2025-01-23
Update segment id

Chris7:
Update segment id to prevent collisions.


dkhofer:
I'll approve this change, but for more context, the reason I didn't put sequence_end on segments is that it can be inferred from sequence_start plus the length of the segment sequence.  So the node ID hash could use the segment length instead of sequence_end and should still be unique.  But not a big deal either way

dkhofer (approved):

PR #134, opened 2025-01-24
Add command to derive a subgraph of a block group

dkhofer:
Given a sample + region + range, creates a new sample with a copy of the region's block group that contains the subgraph that has only nodes and edges that are in the given range, when the range is applied to the current path for the block group.

Some behavior is up for discussion.  Currently does not copy the existing block over to the new sample, but I could be persuaded to add that.  Also doesn't check if the start and end of the range are on nodes that are "articulated", meaning nodes that are necessarily part of every path.  If the node isn't articulated, then edges that are "parallel" to it will be ignored.  I'll address articulated nodes in a followup PR (basically warn the user before dropping "parallel" edges).

This PR is at a good stopping point.  There's other functionality that I'll be implementing in the followup PR, including "chunks" (multiple subgraphs in one command) and some other related command-line flags.

Example usage:

```
gen import --fasta fixtures/simple.fa
gen update --fasta fixtures/aa.fa --start 3 --end 5 --region-name m123 --new-sample test2
gen derive-subgraph --sample test2 --new-sample test3 --region m123:2-10
```


src/imports/gfa.rs
```diff
@@ -78,6 +83,35 @@ pub fn import_gfa<'a>(
     }
     bar.finish();
 
+    let pure_source_refs = source_refs_in_links
```
dkhofer, line 86: The code cleanup elsewhere required adding links from the virtual source node and to the virtual target node from segments here, as needed

src/graph_operators.rs
```diff
@@ -0,0 +1,70 @@
...
+            parent_block_group_id = block_group.id;
+            new_block_group_id = new_bg_id;
+        }
+    }
+
+    if new_block_group_id == 0 {
+        panic!("No region found with name: {}", region_name);
+    }
+
+    BlockGroup::clone_subgraph(
```
Chris7, line 44: maybe instead of clone call this slice_graph or something? clone imo implies it'll be the same.

src/models/block_group.rs
```diff
@@ -757,6 +759,134 @@ impl BlockGroup {
...
+            .nodes()
+            .find(|node| {
+                node.node_id == start_block.node_id
+                    && node.sequence_start == start_block.sequence_start
+            })
+            .unwrap();
+        let end_node = current_graph
+            .nodes()
+            .find(|node| {
+                node.node_id == end_block.node_id && node.sequence_start == end_block.sequence_start
```
Chris7, line 793: i'm not sure if this will always work. The block above is based on paths, but the graph blocks may actually diverge from when the path was made, right? I'm wondering if it actually needs to see if the end node contain the sequence_start vs. is equal to it.

src/models/block_group.rs
```diff
@@ -757,6 +759,134 @@ impl BlockGroup {
...
+            .collect::<HashSet<_>>();
+        let mut new_path_edge_ids = vec![];
+        new_path_edge_ids.push(new_start_edge.id);
+        for current_edge in current_edges {
+            if new_edge_id_set.contains(&current_edge.id) {
+                new_path_edge_ids.push(current_edge.id);
+            }
+        }
+        new_path_edge_ids.push(new_end_edge.id);
+        Path::create(
```
Chris7, line 883: do we want this function to always make a new path?

src/models/block_group.rs
```diff
@@ -757,6 +759,134 @@ impl BlockGroup {
         );
         paths[0].clone()
     }
+
+    pub fn clone_subgraph(
```
Chris7, line 763: since this is path specific, should this function actually be on the path?

Chris7: or maybe what we'd want is this function to make a subgraph between 2 nodes, and other places can do whatever they need to find the start/end nodes and pass it though.

src/models/block_group.rs
```diff
@@ -757,6 +759,134 @@ impl BlockGroup {
...
+        let subgraph_edges = all_intermediate_edges(&current_graph, start_node, end_node);
+
+        // Filter out internal edges (boundary edges) that don't exist in the database
+        let subgraph_edge_ids = subgraph_edges
+            .iter()
+            .map(|(_to, _from, edge_info)| edge_info.edge_id)
+            .collect::<Vec<_>>();
+        let source_edges = Edge::bulk_load(conn, &subgraph_edge_ids);
+
+        let source_block_group_edges =
```
Chris7, line 805: should we make a function for this? We could be loading millions of edges in order to grab 5-6 that are in subgraph_edge_ids

src/models/block_group.rs
Chris7: i think that would be the best setup here. I understand this uses a path to pinpoint locations, and grabs whatever happens to be within that so this probably does belong here. I think it would be a generally useful thing to have this not be linked to paths, but paths can be one way to get the nodes to select.

src/models/block_group.rs
Chris7: it's not obvious to me we'd always want a new path, especially a subselection of the data and defaulting it to the path name that could be like `chr1`.

Chris7 (approved):
Nothing blocking, though i think it would be better to make the blockgroup part node based and have the part to select which nodes to use be separated.

src/graph_operators.rs
dkhofer: I changed it to `derive_subgraph` to be in line with the actual command name

src/models/block_group.rs
dkhofer: For the purposes I'm aware of, yes, but I moved the new path creation up into the derive subgraph code, out of here

src/models/block_group.rs
dkhofer: Yep, I changed it to work with nodes instead of a path, and moved all the path stuff up into the derive subgraph code.

src/models/block_group.rs
dkhofer: Good point, done

PR #136, opened 2025-01-29
Import combinatorial library

dkhofer:


src/imports/library.rs
```diff
@@ -0,0 +1,274 @@
...
+        let name = String::from_utf8(record.name().to_vec()).unwrap();
+        let seq = Sequence::new()
+            .sequence_type("DNA")
+            .sequence(&sequence)
+            .save(conn);
+        let node_id = Node::create(
+            conn,
+            &seq.hash,
+            calculate_hash(&format!(
+                "{region_name}:{ref_start}-{ref_end}->{sequence_hash}",
```
Chris7, line 63: we should probably put the record name in here as well. For example, suppose we had these 2 entries:
```
>start_codon
ATG
>atg_but_not_start_codon
ATG
```

We'd have them both be the same node and used in that manner.

Chris7 (approved):
Looks good, just 1 comment on node hash.

src/imports/library.rs
dkhofer: :+1:, done

PR #140, opened 2025-02-06
Refactor operations

Chris7:
This makes files for operation a join table to support multiple files. Also removed the unused tables in core.


dkhofer (approved):
Thanks for the cleanup work.  LGTM

PR #142, opened 2025-02-10
Translate path node to current graph.

Chris7:
This translates a path nodes to the current graph in gfa exports.

### DIFF
--- a/src/exports/gfa.rs
+++ b/src/exports/gfa.rs
@@ -1,4 +1,5 @@
 use crate::gfa::{path_line, write_links, write_segments, Link, Path as GFAPath, Segment};
+use crate::models::sequence::Sequence;
 use crate::models::{
     block_group::BlockGroup,
     block_group_edge::BlockGroupEdge,
@@ -62,10 +63,10 @@ pub fn export_gfa(
     let file = File::create(filename).unwrap();
     let mut writer = BufWriter::new(file);
 
-    let mut segments = vec![];
+    let mut segments = BTreeSet::new();
     for block in &blocks {
         if !Node::is_terminal(block.node_id) {
-            segments.push(Segment {
+            segments.insert(Segment {
                 sequence: block.sequence(),
                 node_id: block.node_id,
                 sequence_start: block.start,
@@ -76,7 +77,6 @@ pub fn export_gfa(
             });
         }
     }
-    write_segments(&mut writer, &segments);
 
     let mut links = BTreeSet::new();
     for (source, target, edge_info) in graph.all_edges() {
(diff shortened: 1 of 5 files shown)


src/exports/gfa.rs
```diff
@@ -62,10 +59,10 @@ pub fn export_gfa(
     let file = File::create(filename).unwrap();
     let mut writer = BufWriter::new(file);
 
-    let mut segments = vec![];
+    let mut segments = BTreeSet::new();
```
Chris7, line 62: i kept this, just seemed better to protect against duplicates

src/exports/gfa.rs
```diff
@@ -104,115 +100,138 @@ pub fn export_gfa(
...
+    struct PathNode {
+        node: (GraphNode, Strand),
+        // path_index serves as a pointer at which path block the added node was at. This is because
+        // in a DFS, we may reset to a previous position for a search and need to reset the position
+        // in the path as well.
+        path_index: usize,
+        prev: Option<Rc<PathNode>>,
+    }
+
+    let start_nodes = graph
```
dkhofer, line 133: There should just be one start node, right?

src/imports/gfa.rs
```diff
@@ -229,10 +220,24 @@ pub fn import_gfa<'a>(
...
+                    chromosome_index: 0,
+                    phased: 0,
+                })
+                .collect::<Vec<BlockGroupEdgeData>>(),
+        );
         Path::create(conn, path_name, block_group.id, &path_edge_ids);
     }
 
-    for input_walk in &gfa.walk {
+    for (walk_index, input_walk) in (1..).zip(&gfa.walk) {
```
dkhofer, line 240: Using zip is fine, but why not enumerate?

src/imports/gfa.rs
```diff
@@ -229,10 +220,24 @@ pub fn import_gfa<'a>(
...
+        created_blockgroup_edges.extend(path_edge_ids.iter());
+
+        BlockGroupEdge::bulk_create(
+            conn,
+            &path_edge_ids
+                .iter()
+                .map(|id| BlockGroupEdgeData {
+                    block_group_id: block_group.id,
+                    edge_id: *id,
+                    chromosome_index: 0,
```
dkhofer, line 232: Seems like it would make sense to enumerate paths and use that index here?  I guess you'd need to ensure the walk index below is different

src/exports/gfa.rs
```diff
@@ -104,115 +100,138 @@ pub fn export_gfa(
...
-    node_ids.push(format!(
-        "{}.{}.{}",
-        end_block.node_id, end_block.start, end_block.end
-    ));
-
-    node_ids
-}
-
-fn get_path_links(
+fn translate_path_links(
```
dkhofer, line 109: This method looks reasonable to me, but could you add a couple unit tests confirming it works?

src/exports/gfa.rs
Chris7: yeah, i got rid of the loop since it should always be our source

src/imports/gfa.rs
Chris7: so you can't start enumerate at 1, this is the way to do that. It's to prevent walks from interfering w/ the paths above which get index 0

src/imports/gfa.rs
Chris7: I think each path should be index 0. I'm kind of mentally treating a path as the linear reference genome (0 index) and walks as alternative paths

src/exports/gfa.rs
Chris7: yeah....i should :). I moved to to graphs as well and made it slightly more genericish.

PR #146, opened 2025-02-13
Cycle detection

Chris7:
Handling of cycles in GFA import:

If a cycle is detected, we break it by identify the cycle, then rotating the cycle so it starts with the lowest node id. A start node is then inserted before this node id and an end node is appended to the list. This is the new linear version of the cycle. An edge is made between the end node -> start node just so we can detect cycles easier in the future.


Chris7:
@bobvh @dkhofer Strawman for handling cycles. This introduces a node dedicated as a cycle. A circle would connect to this element (vs. linearizing it to a start/end node).

Some thoughts:
* Graphs with cycles within them, for example a lasso shape. Should we just ignore this? There won't be a biological entity with this shape, but you could think about a linearization of a plasmid as being this kind of shape where both variants can exist. Doesn't seem worth it to me.
* The way strongly connected components (SCC) works is a bit wrong here. It only adds elements to a single SCC, so if there are multiple entrypoints to a cycle, it won't show up (see https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm, node 4.4 for example will be excluded from the upper left SCC). 
* We'd need to rework a lot of our path finding code that relies on start/end being separate points.

bobvh:
> @bobvh @dkhofer Strawman for handling cycles. This introduces a node dedicated as a cycle. A circle would connect to this element (vs. linearizing it to a start/end node).

Breaking the DAG model is going to complicate more than just path finding. Could we have a third dummy node instead?
So you'd have start/end/circle as dummy nodes. Start is always the source, but the sink can be either end or circle (or whatever name).

That way things stay easier internally, and on export to GFA we can introduce a final edge if a the graph ends in a "circle" dummy.

On the lasso's: there are so many weird peptide modifications out there that I think they are better handled on the sequence model than in the block group. The field as a whole hasn't agreed on a standard representation, but if we have something in the sequence parsing code that considers something like \lasso-entry\ or \5mC\ (methylcytosine) as a single character, that could be interesting.

Chris7:
Doing an edge from End -> Start makes it kind of look pretty good
image

bobvh:
In the terminal viewer I could make it work for hierarchical layout by wrapping around (with an origin indicator so users don't keep scrolling forever).

src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub mod operation_management;
 pub mod patch;
 mod progress_bar;
 pub mod range;
-#[cfg(test)]
+// #[cfg(test)]
```
dkhofer, line 22: What's going on here?

src/imports/gfa.rs
```diff
@@ -321,6 +326,86 @@ pub fn import_gfa<'a>(
...
+                if !has_start && Node::is_start_node(node.node_id) {
+                    has_start = true;
+                } else if !has_end && Node::is_end_node(node.node_id) {
+                    has_end = true;
+                };
+                if has_start && has_end {
+                    break;
+                }
+            }
+            // honestly not sure what to do about graphs where we have just one terminal
```
dkhofer, line 356: I think it should be an error.  Is there some case you can think of where it might be valid?

src/imports/gfa.rs
```diff
@@ -321,6 +326,86 @@ pub fn import_gfa<'a>(
...
+            if !has_start && !has_end {
+                // from the subgraph, we want to find a deterministic sort of ordered elements.
+                // Kosaraju returns nodes in arbitrary order. We use DFS and then rotate the vector
+                // so the first node_id starts the list for consistency.
```
dkhofer, line 360: Would it make sense to also check the walks and paths, and use the segments/nodes that occur at the start of those?

dkhofer: Ideally they'd be the same as the min node ID, not sure what happens in reality

src/lib.rs
Chris7: just so i can use the save_graph function outside of test builds

src/imports/gfa.rs
Chris7: oh that's a really good point

src/imports/gfa.rs
Chris7: There are such things as lasso peptides
image

src/imports/gfa.rs
Chris7: i did this, but it ended up being already setup due to the way walks/paths are added already with start/end nodes. This causes the cycle detection code to skip these cycles since there is already a start/end.

PR #147, opened 2025-02-13
Add operation to derive chunks

dkhofer:
Example usage:

```
cargo run -- init
cargo run -- import --fasta fixtures/simple.fa
# Create two example integration points
cargo run -- update --fasta fixtures/aa.fa --start 3 --end 5 --region-name m123 --new-sample test1
cargo run -- update --fasta fixtures/aa.fa --start 15 --end 20 --region-name m123 --sample test1 --new-sample test2
# Derive chunks with breakpoints in between the edits
cargo run -- derive-chunks --new-sample splits --region m123 --breakpoints 1,8,25 --sample test2
# View one of the new chunks
cargo run -- view -s splits m123.2
```

### DIFF
--- a/src/graph_operators.rs
+++ b/src/graph_operators.rs
@@ -263,6 +263,7 @@ fn get_parent_block_group_id(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::imports::fasta::import_fasta;
     use crate::models::{
         block_group_edge::BlockGroupEdgeData, collection::Collection, edge::Edge, metadata,
         node::Node, operations::setup_db, sequence::Sequence, strand::Strand,
@@ -270,6 +271,8 @@ mod tests {
     use crate::test_helpers::{
         get_connection, get_operation_connection, setup_block_group, setup_gen_dir,
     };
+    use crate::updates::fasta::update_with_fasta;
+    use std::path::PathBuf;
 
     #[test]
     fn test_derive_chunks_one_insertion() {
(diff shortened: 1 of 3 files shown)


src/graph_operators.rs
```diff
@@ -15,102 +15,213 @@ use rusqlite::Connection;
...
+        start: range_start,
+        end: path_length,
+    });
+
+    chunk_ranges
+}
+
 #[allow(clippy::too_many_arguments)]
-pub fn derive_subgraph(
+pub fn derive_chunks(
```
dkhofer, line 109: Most of the changes to this method are just generalizing from one subgraph/chunk to many

src/models/block_group.rs
```diff
@@ -830,7 +850,7 @@ impl BlockGroup {
         let new_start_edge = Edge::create(
             conn,
             PATH_START_NODE_ID,
-            -1,
+            0,
```
dkhofer, line 853: Using -1 was the wrong choice, it prevents edges from the start node from showing up when the graph is loaded.  I have a to-do to look through the code and see where else this is being done

src/graph_operators.rs
```diff
@@ -15,102 +15,217 @@ use rusqlite::Connection;
...
+// Given specific points on a path, returns a list of ranges that cover the entire path, broken up
+// by the input points
+pub fn get_breakpoint_ranges(
+    conn: &Connection,
+    collection_name: &str,
+    parent_sample_name: Option<&str>,
+    new_sample_name: &str,
+    region_name: &str,
+    backbone: Option<&str>,
+    breakpoints: &str,
```
Chris7, line 80: is there a reason this is a string vs. a vec here?

src/graph_operators.rs
```diff
@@ -15,102 +15,217 @@ use rusqlite::Connection;
...
+            end: breakpoint,
+        });
+        range_start = breakpoint;
+    }
+    chunk_ranges.push(Range {
+        start: range_start,
+        end: path_length,
+    });
+
+    chunk_ranges
```
Chris7, line 109: is this supposed to be sorted in some way? or a check that the breakpoints are within the length of the path?

Chris7: it seems like this and the above are pretty similar, should these be merged to one function?

src/graph_operators.rs
```diff
@@ -15,102 +15,217 @@ use rusqlite::Connection;
...
+            new_block_group_name.as_str(),
+        );
+        let new_block_group_id = new_block_group.id;
+
+        let start_coordinate = chunk_range.start;
+        let end_coordinate = chunk_range.end;
+        if (start_coordinate < 0 || start_coordinate > current_path_length)
+            || (end_coordinate < 0 || end_coordinate > current_path_length)
+        {
+            panic!(
```
Chris7, line 149: since you are using result here, can you define error functions for these? It makes it so we can test the code better and handle errors better by delegating the message/etc. to whatever invoked the function.

Chris7 (approved):
Just a few comments. I think the range functions are pretty similar but have different behaviors around the edges, which seems like it might make sense to unify them. For errors, I think we need to start using the errors defined more vs. panics (you can see the stuff like FastaError and what I just did in https://github.com/ginkgobioworks/gen/pull/150). It lets us test the code more effectively, sets us up better for library code, and lets various callers handle the errors in the way that makes sense.

src/graph_operators.rs
dkhofer: No, and vec is better.  I moved the parsing into main.rs

src/graph_operators.rs
dkhofer: I reworked a bunch of it and moved the range calculation into main.rs, it was mostly straightforward and I think it's better to have it there

src/graph_operators.rs
dkhofer: Good point, done.  I have a to-do to do that wherever else it's needed

PR #148, opened 2025-02-15
Add make stitch operation

dkhofer:
Add "make stitch" operation.

```
cargo run -- init
cargo run -- import --fasta fixtures/simple.fa
# Create two example integration points
cargo run -- update --fasta fixtures/aa.fa --start 3 --end 5 --region-name m123 --new-sample test1
cargo run -- update --fasta fixtures/aa.fa --start 15 --end 20 --region-name m123 --sample test1 --new-sample test2
# Derive chunks with breakpoints in between the edits
cargo run -- derive-chunks --new-sample splits --region m123 --breakpoints 1,8,25 --sample test2
# View one of the new chunks
cargo run -- view -s splits m123.2
# Stitch two chunks together
cargo run -- make-stitch --sample splits --new-sample stitched --regions m123.2,m123.3 --new-region m123.2-3
```

### DIFF
--- a/src/graph_operators.rs
+++ b/src/graph_operators.rs
@@ -19,7 +19,7 @@ use std::collections::{HashMap, HashSet};
 use thiserror::Error;
 
 #[derive(Debug, Error, PartialEq)]
-pub enum DeriveGraphError {
+pub enum GraphOperationError {
     #[error("Operation Error: {0}")]
     OperationError(#[from] OperationError),
     #[error("Invalid coordinate(s): {0}")]
@@ -36,13 +36,13 @@ pub fn get_path(
     sample_name: Option<&str>,
     region_name: &str,
     backbone: Option<&str>,
-) -> Result<Path, DeriveGraphError> {
+) -> Result<Path, GraphOperationError> {
     let block_group_id = get_block_group_id(conn, collection_name, sample_name, region_name)?;
 
     if let Some(backbone) = backbone {
         let path = BlockGroup::get_path_by_name(conn, block_group_id, backbone);
         if path.is_none() {
-            return Err(DeriveGraphError::PathNotFound(format!(
+            return Err(GraphOperationError::PathNotFound(format!(
                 "No path found with name {}",
                 backbone
             )));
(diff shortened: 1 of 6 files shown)


src/models/block_group.rs
```diff
@@ -855,6 +858,95 @@ impl BlockGroup {
             .collect::<Vec<_>>();
         let source_edges = Edge::bulk_load(conn, &subgraph_edge_ids);
 
+        // Instead of reusing the existing edges, we create copies of the nodes and edges.  This
```
dkhofer, line 861: This seemed like the easiest solution.  From talking with Bob about it today, one problem is propagating annotations.  If nodes change from one sample to a child sample, but represent the same part of a sequence, it's harder to pass annotations along.  I'm open to other ideas for what to do

Chris7: what are the challenges hit when reusing nodes?

src/graph_operators.rs
```diff
@@ -113,53 +116,87 @@ pub fn derive_chunks(
...
-                })
-                .unwrap();
-            new_path_edge_ids.push(new_end_edge.edge.id);
-        }
+
+        let new_end_source_node_id = new_node_ids_by_old.get(&end_block.node_id).unwrap();
+        let new_end_edge = child_block_group_edges
+            .iter()
+            .find(|e| {
+                e.edge.target_node_id == PATH_END_NODE_ID
```
Chris7, line 193: can you use the Node::is_start_node/etc. methods?

src/graph_operators.rs
```diff
@@ -208,12 +245,256 @@ fn get_block_group_id(
...
+            }
+        } else if !path_edge.is_start_edge() {
+            new_path_edge_ids.push(path_edge.id);
+        }
+    }
+
+    new_path_edge_ids.push(concatenated_path_edges[concatenated_path_edges.len() - 1].id);
+
+    // Part 5: Create bg edges for the new edges
+    let mut chromosome_index_counter = edges_to_reuse
```
Chris7, line 435: since we're adding +1 here, it seems like we're trying to get the first unused chromosome_index. If  this is true, shouldn't this by max_by?

Chris7 (approved):
Thanks for the info on reusing node ids. non-blocking comments added.

src/models/block_group.rs
dkhofer: Yep, done

src/graph_operators.rs
dkhofer: Oops, yep, done

src/graph_operators.rs
dkhofer: 🤦 No idea why I had min_by there, fixed

PR #149, opened 2025-02-18
Terminal viewer partitions

bobvh:
This resolves the issue of recursive layout algorithms overflowing the stack on large pangenomes. It required a rather big refactoring, but it's in a much better state now and more compatible with our internal structs. The way I chose to solve it is:
- find articulation points (GraphNodes that every possible path must pass through)
     `fn find_articulation_points(graph: &DiGraphMap<GraphNode, GraphEdge>) -> Vec<GraphNode>`
- partition the graph in non-overlapping subgraphs, in the StableGraph format the Sugiyama crate needs
    - iterate over a topological sorting of the nodes, and every time you encounter an articulation point consider splitting it by counting how many nodes we've been accumulating so far
    - cut edges are recorded for reassembly later
    `struct Partition { parts: Vec<SugiyamaGraph>, inter_part_edges: HashMap<(usize, usize), Vec<PartitionEdge>>,}`
- compute a BaseLayout for the partition subgraphs around the viewport window, this pretends every node is a fixed size 
- transform into a ScaledLayout that does take into account zoom level, sequences, etc.
    - this way we don't have the compute each layout more than once

There's also a flag now to go directly to the coordinate you're interested in:`gen view -p node_id:position`. (that's in proper Nodes, not GraphNodes). I did cut back on the colors etc. for now, it turns out the default Apple terminal emulator doesn't support much fancy, so we need to be able to fall back to something simple when detecting that.


src/views/block_group.rs
```diff
@@ -25,19 +19,20 @@ use ratatui::{
...
 
 pub fn view_block_group(
     conn: &Connection,
     name: &str,
     sample_name: Option<String>,
     collection_name: &str,
+    position: Option<String>, // Node ID and offset
 ) -> Result<(), Box<dyn Error>> {
     // Get the block group for two cases: with and without a sample
     let block_group = if let Some(ref sample_name) = sample_name {
```
Chris7, line 31: do you actually need Some(ref here?

src/views/block_group.rs
```diff
@@ -50,127 +45,39 @@ pub fn view_block_group(
...
-    // Confirm that there is only one layout, which means that the graph is connected
-    assert_eq!(layouts.len(), 1);
-
-    // Store that one layout and convert the block ids from usize to u32
-    let layout: Vec<(u32, (f64, f64))> = layouts[0]
-        .0
-        .iter()
-        .map(|(id, (x, y))| (*id as u32, (*x, *y)))
-        .collect();
+    // Get the node object corresponding to a node id
```
Chris7, line 48: would we have an interface for paths/accessions here too?

src/views/block_group.rs
```diff
@@ -392,6 +298,4 @@ pub fn view_block_group(
 }
 
 #[cfg(test)]
-mod tests {
-    //use super::*;
-}
+mod tests {}
```
Chris7, line 301: should just delete this if there are no tests

src/views/block_group_viewer.rs
```diff
@@ -1,26 +1,160 @@
...
+        let base_layout = BaseLayout::with_origin(block_graph, origin.clone());
+
+        // Get the sequences for the nodes in the base_layout subgraph
+        let node_sequences = Node::get_sequences_by_node_ids(
+            conn,
+            &base_layout
+                .layout_graph
+                .nodes()
+                .map(|node| node.node_id)
+                .collect::<std::collections::BTreeSet<i64>>()
```
Chris7, line 105: do we care about sequence order here (BTreeSet vs HashSet)?

src/views/block_group_viewer.rs
```diff
@@ -93,41 +238,111 @@ impl Viewer {
...
                 // Print the labels
-                for (block_id, (label, x, y)) in &self.layout.labels {
+                for (block, ((x, y), (x2, _y2))) in self.scaled_layout.labels.iter() {
                     // Skip labels that are not in the visible area (vertical)
                     if (*y as i32) < viewport_bottom || (*y as i32) >= viewport_top {
                         continue;
                     }
+
+                    // Handle dummy nodes (start and end) differently than other nodes
+                    if block.node_id == node::PATH_START_NODE_ID
```
Chris7, line 253: can you use Node::is_start_end and is_end_node here?

src/views/block_group_viewer.rs
```diff
@@ -1,26 +1,160 @@
...
+
+        // Create initial scroll state with origin block selected (or first non-start/end node if origin is start)
+        let mut scroll_state = ScrollState::default();
+        if let Some(origin) = origin_block {
+            if origin.node_id == node::PATH_START_NODE_ID {
+                // Find the first non-start/end node by looking at outgoing neighbors of the start node
+                scroll_state.selected_block = base_layout
+                    .layout_graph
+                    .neighbors_directed(origin, Direction::Outgoing)
+                    .find(|node| node.node_id != node::PATH_END_NODE_ID);
```
Chris7, line 129: should use the helper functions in Node for these comparisons

src/views/block_group_viewer.rs
```diff
@@ -326,6 +624,80 @@ pub fn clip_line(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::graph::{GraphEdge, GraphNode};
+    use crate::models::strand::Strand;
+    use petgraph::graphmap::DiGraphMap;
+
+    #[allow(dead_code)]
+    fn make_test_graph(
```
Chris7, line 632: any reason to keep this?

src/views/block_layout.rs
```diff
@@ -1,368 +1,1195 @@
...
+                },
+            )
+        }))
+    }
+
+    #[test]
+    fn test_make_test_graph() {
+        let edges = vec![(0, 1), (1, 2), (2, 3), (3, 4), (4, 5)];
+        let graph = make_test_graph(edges, None);
+        println!("Graph: {:#?}", graph);
```
Chris7, line 810: should remove excess prints

Chris7 (approved):
Just minor stuff on using the Node functions for start/end. Overall looks good and an improvement over what it was.

src/views/block_group.rs
bobvh: Yes, I borrow sample_name again a few lines down in an error message, and without the ref the compiler complains about a borrow after move.

src/views/block_group.rs
bobvh: That's a good idea, people are going to want to see their gene of interest highlighted after moving to that general area.

src/views/block_group_viewer.rs
bobvh: I thought it was going to be easier for get_sequences_by_node_ids to process many ids if I retained the order on the node_ids, but I haven't measured the difference. Seems to work fine with HashSet, so I changed it over.

src/views/block_group_viewer.rs
bobvh: Neat. Changed there and in a few other spots.

src/views/block_group_viewer.rs
bobvh: The tests do require it, but Clippy doesn't consider that as a "use". That little helper function did get bigger over time, so I'd be okay with it living somewhere else. But that still would need something to avoid clippy complaining, no?

PR #150, opened 2025-02-18
Check if position is beyond path length

Chris7:
Found this problem when i applied a vcf to the wrong fasta. This ensures we aren't making changes outside the scope of the interval tree. I put it in setup_up_new_edges as a general catch area, but also in the vcf update code where i initially found the bug so it can bail early before parsing a few gigs of vcf.

### DIFF
--- a/src/updates/vcf.rs
+++ b/src/updates/vcf.rs
@@ -173,6 +173,8 @@ struct VcfEntry {
 pub enum VcfError {
     #[error("Operation Error: {0}")]
     OperationError(#[from] OperationError),
+    #[error("Invalid Record: {0}")]
+    InvalidRecord(String),
 }
 
 pub fn update_with_vcf<'a>(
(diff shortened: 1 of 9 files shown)


src/updates/vcf.rs
```diff
@@ -289,6 +293,17 @@ pub fn update_with_vcf<'a>(
                         };
                         let sample_path =
                             PathCache::lookup(&mut path_cache, sample_bg_id, seq_name.clone());
+                        let path_length = if let Some(l) = path_lengths.get(&sample_path.id) {
```
dkhofer, line 296: Is it not possible to do this?

```
let path_length = path_lengths.entry(&sample_path.id).or_insert(sample_path.length(conn));
```

dkhofer (approved):
One suggestion for better code, but LGTM!

src/updates/vcf.rs
Chris7: probably is, i'll try :)

PR #152, opened 2025-02-21
Integrate graph view into operations

Chris7:


src/views/operations.rs
```diff
@@ -36,15 +41,31 @@ struct OperationRow<'a> {
     summary: OperationSummary,
 }
 
-pub fn view_operations(conn: &Connection, operations: &[Operation]) -> Result<(), io::Error> {
+fn restore_terminal() {
```
bobvh, line 44: Have you confirmed that this actually does something if you use the Ratatui init?

src/views/operations.rs
```diff
@@ -68,7 +89,18 @@ pub fn view_operations(conn: &Connection, operations: &[Operation]) -> Result<()
     let mut terminal = Terminal::new(backend)?;
```
bobvh, line 89: Here's where  ` let mut terminal = ratatui::init();` could come in.

src/views/operations.rs
bobvh: Stack trace saving would have to be a separate hook, so I don't know if there's any benefit from combining approaches.

PR #154, opened 2025-02-25
Blockgroup browser

bobvh:


src/views/collection.rs
```diff
@@ -0,0 +1,695 @@
...
+    use rusqlite::Connection;
+
+    /// For these tests we create an in-memory database, run minimal schema
+    /// creation, and insert data to test gather_collection_explorer_data.
+    #[test]
+    fn test_gather_collection_explorer_data() {
+        // 1) Set up an in-memory database
+        let conn = Connection::open_in_memory().unwrap();
+
+        // Minimal schema for the required tables, adapted from migrations
```
Chris7, line 556: why not use the test harness for an in memory db?

src/views/block_group.rs
```diff
@@ -80,86 +87,185 @@ pub fn view_block_group(
...
+                sidebar_layout[1]
             };
-            let status_bar_area = outer_layout[1];
 
-            let status_message = format!(
-                "{message} | return: show information on block | q=quit",
-                message = Viewer::get_status_line()
-            );
+            // Sidebar
+            explorer_state.has_focus = focus_zone == "sidebar";
```
Chris7, line 156: in other places i have the main interface (like this), holding focus state. I'm ok with widgets knowing it as well, but we should coordinate somehow.

src/views/block_group.rs
```diff
@@ -173,35 +279,88 @@ pub fn view_block_group(
...
+                        KeyCode::Char('q') => break,
                         KeyCode::Tab => {
-                            // Future implementation: switch between panels
-                        }
-                        KeyCode::BackTab => {
-                            // Future implementation: switch between panels
-                        }
-                        KeyCode::Esc => {
-                            show_panel = false;
+                            if key.modifiers == KeyModifiers::SHIFT {
```
Chris7, line 305: this actually won't work, you need to use BackTab

src/views/block_group.rs
```diff
@@ -173,35 +279,88 @@ pub fn view_block_group(
...
+                                // Shift+Tab - cycle backwards
```
Chris7, line 306: we need some better focus thing to handle these. mine is also a bit of a mess.

src/views/block_group_viewer.rs
```diff
@@ -18,6 +18,8 @@ use ratatui::{
 use rusqlite::Connection;
 use std::collections::{HashMap, HashSet};
 
+// Show more visual information if DEBUG
+const DEBUG: bool = false;
```
Chris7, line 22: should use the logger env i guess instead of this.

src/views/block_group.rs
```diff
@@ -80,86 +87,185 @@ pub fn view_block_group(
...
     let mut last_tick = Instant::now();
     let mut show_panel = false;
+    let show_sidebar = true;
     let mut tui_layout_change = false;
+
+    // Focus management
+    let mut focus_zone = "canvas";
+
+    // Create explorer and its state that persists across frames
+    let mut explorer = CollectionExplorer::new(conn, collection_name);
```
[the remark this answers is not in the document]
bobvh: I consider the Explorer sidebar as its own widget that lives independently from the Viewer. That way, I don't force it on the operations UI, for example. I also need to keep the state on the listview third-party widget, so I refer to that state from inside of the explorer state.

src/views/block_group.rs
bobvh: You're right, and that's the second time I've run into that. Fixed

src/views/block_group.rs
bobvh: Yeah I saw the index based focus cycler you had in operations but it didn't fit well here. I actually made the sidebar permanent because of the focus cycling. Fyi: I didn't use ctrl+arrows because the window manager on macOS uses that.

src/views/block_group_viewer.rs
bobvh: Which one? Check if RUST_LOG is "debug" as equivalent?

src/views/collection.rs
```diff
@@ -0,0 +1,695 @@
...
+        sample_block_groups.insert(sample.clone(), pairs);
+    }
+
+    // 5) Direct "nested" collections: must start with "full_collection_name + /" but no further delimiter
+    let direct_prefix = format!("{}{}", full_collection_name, "/");
+
+    let sibling_candidates = Collection::query(
+        conn,
+        "SELECT * FROM collections
+         WHERE name GLOB ?1",
```
Chris7, line 136: why this over like?

src/views/collection.rs
```diff
@@ -0,0 +1,695 @@
...
+        let items = self.get_display_items(state);
+        if items.is_empty() {
+            return;
+        }
+
+        let current_idx = state.list_state.selected.unwrap_or(0);
+        state.list_state.selected = self.find_prev_selectable(state, current_idx);
+    }
+
+    pub fn handle_input(&self, state: &mut CollectionExplorerState, key: KeyEvent) {
```
Chris7, line 332: why isn't this using the stateful widget vs. the custom state?

src/views/collection.rs
Chris7: it is, we need to condense it, but put this at the top:

```
setup_gen_dir();
        let conn = &mut get_connection(None);
        let db_uuid = get_db_uuid(conn);
        let operation_conn = &get_operation_connection(None);
        setup_db(operation_conn, &db_uuid);
```

src/views/collection.rs
bobvh: We're emulating file paths, so it seemed appropriate. (not a strong opinion though)

src/views/collection.rs
bobvh: I wanted to be able to skip the headers when scrolling down the sidebar, that's not a part of tui-widget-list currently.

PR #155, opened 2025-03-03
Viewer - handle synchronization to external changes

bobvh:
In the final commit to my previous PR I had added a connection refreshing routine that opened and closed the db connection. That was overkill and by using Connection::open instead of get_connection it introduced a new bug (Connection::open does not load he extensions that other methods rely on). Now we're not changing the conn at all.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -589,11 +589,9 @@ fn main() {
             collection,
             position,
         }) => {
-            conn.execute("BEGIN TRANSACTION", []).unwrap();
             let collection_name = &collection
                 .clone()
                 .unwrap_or_else(|| get_default_collection(&operation_conn));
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -593,15 +593,16 @@ fn main() {
             let collection_name = &collection
                 .clone()
                 .unwrap_or_else(|| get_default_collection(&operation_conn));
+            conn.execute("END TRANSACTION", []).unwrap();
```
Chris7, line 596: you don't need a transction at all here, transactions are only really for writing data

Chris7 (commented):
I think the issue is the use of transactions at all. For read operations, don't bother w/ them.

